PR #11, opened 2023-06-08
Populate python

Nazim-crim:
Replacing populate.sh with a python script


dchandan:
I think this is outdated now. Should we delete this?

Nazim-crim:
@dchandan yes we can

PR #16, opened 2023-08-24
Further revision to the new ingestion workflow

dchandan:
This PR makes further revision to the ingestion workflow by incorporating @huard's recent PR from another branch #14. Additional small architecture changes are included.


dchandan:
Not sure why I can't add @huard as a reviewer. David, please review this PR.

STACpopulator/input.py
```diff
@@ -58,21 +63,47 @@ def reset(self):
...
                 self.catalog_head = ref.follow()
                 self._depth -= 1
                 yield from self
 
+    def extract_metadata(self, ds: siphon.catalog.Dataset) -> MutableMapping[str, Any]:
+        # Get URL for NCML service
+        url = ds.access_urls["NCML"]
+
+        LOGGER.info("Requesting NcML dataset description")
+        r = requests.get(url)
```
huard, line 84: I've found that if you use this
```r = requests.get(url, params={"catalog": catalog, "dataset": dataset})```

You get more information in the response, in particular, you get a group called "THREDDSMetadata" with all the services offered by THREDDS.  The line below (access_urls) would not be needed.

File: STACpopulator/input.py [hunk 2]
```diff
@@ -58,21 +63,47 @@ def reset(self):
...
+
+        # Write response to temporary file
+        f = NamedTemporaryFile()
+        f.write(r.content)
+
+        # Convert NcML to CF-compliant dictionary
+        attrs = xncml.Dataset(f.name).to_cf_dict()
```
huard, line 91: Note to myself: I could add a `from_xml` class method to `xncml.Dataset` to avoid having to create a temp file.

implementations/CMIP6-UofT/add_CMIP6.py
```diff
@@ -39,11 +40,14 @@ def __init__(
             print(item)
         super().__init__(stac_host, data_loader, config_filename)
 
-    def process_stac_item(self):  # noqa N802
+    def handle_ingestion_error(self, error: str, item_name: str, item_data: MutableMapping[str, Any]):
+        pass
+
+    def create_stac_item(self, item_name: str, item_data: MutableMapping[str, Any]) -> MutableMapping[str, Any]:
```
huard, line 46: In practice, I suspect this method will both ingest and validate in one step, but let's keep that structure.

huard (approved):

STACpopulator/input.py
huard: Try this: 

https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/ncml/birdhouse/testdata/xclim/cmip6/sic_SImon_CCCma-CanESM5_ssp245_r13i1p2f1_2020.nc?catalog=https%3A%2F%2Fpavics.ouranos.ca%2Ftwitcher%2Fows%2Fproxy%2Fthredds%2Fcatalog%2Fbirdhouse%2Ftestdata%2Fxclim%2Fcmip6%2Fcatalog.html&dataset=birdhouse%2Ftestdata%2Fxclim%2Fcmip6%2Fsic_SImon_CCCma-CanESM5_ssp245_r13i1p2f1_2020.nc

huard: It's the link provided on the THREDDS web page for the NcML service.

STACpopulator/input.py
huard: Ok, I'll make a PR for this.

STACpopulator/populator_base.py
```diff
@@ -84,21 +86,33 @@ def ingest(self) -> None:
         if not stac_collection_exists(self.stac_host, self.collection_id):
             LOGGER.info(f"Creating collection '{self.collection_name}'")
             pystac_collection = create_stac_collection(self.collection_id, self._collection_info)
-            post_collection(self.stac_host, pystac_collection)
+            post_stac_collection(self.stac_host, pystac_collection)
```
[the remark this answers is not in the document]
dchandan: I'm not sure if I understood you correctly: what I understand is you are proposing to move the logic around creation of the collection itself to the contractor (or called from the constructor)? I think that's not a bad idea, that logic is separate from the "ingestion" itself.

STACpopulator/populator_base.py
```diff
@@ -84,21 +86,33 @@ def ingest(self) -> None:
...
+                LOGGER.error(f"Validation failed for item {item_name}")
+                self.handle_ingestion_error("Validation Error", item_name, item_data)
+
+    @abstractmethod
+    def handle_ingestion_error(self, error: str, item_name: str, item_data: MutableMapping[str, Any]):
         pass
 
     @abstractmethod
-    def process_stac_item(self):  # noqa N802
+    def create_stac_item(self, item_name: str, item_data: MutableMapping[str, Any]) -> MutableMapping[str, Any]:
```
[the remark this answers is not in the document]
dchandan: Good question. Right now the `create_stac_collection` is in `stac_utils` because it does not contain any logic that might need to be overridden by any derived classes of `STACpopulatorBase`. But, that is not to say that in future there might not be a need to override some parts of the logic. Do you think we should move all the create/post stac item/collection functions to `STACpopulatorBase`?

STACpopulator/input.py
huard: https://github.com/xarray-contrib/xncml/pull/54

STACpopulator/populator_base.py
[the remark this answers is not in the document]
Nazim-crim: Not necessarily, I think there should be a different file for all the code where `requests` are sent to the stac host. In the future `stac_utils` might become too big and have too many different responsabilies which would hinder the code readability. I would just keep `url_validate` inside `stac_utils`. Also, the `create_stac_collection` could be in `STACpopulatorBase` as a non-abstract method and be changed later on if there's a need for that. Let me know what you think.

STACpopulator/api_requests.py
```diff
@@ -0,0 +1,51 @@
+import os
```
Nazim-crim, line 1: add `class bcolors`  that is in .depecrated/collection_processor.py since it has not been defined and is used in `post_stac_collection`

dchandan:
@huard @Nazim-crim I wanted to update you both that I have some un-pushed further development on this branch. I am on vacation right now, so will push it after I am back next week :)

huard:
We were planning to do a sprint today and tomorrow on this. It would be useful to know what you did already and what remains to be done.

dchandan:
Writing a quick response here as the internet connection in my current hotel is extremely bad. The main changes i made is to the pydantic data model itself. I now have this working well and the serialization to JSON also works well (I had mentioned this was a problem earlier). I've tested the functionality to post STAC item and it works well, as far as I can say. There are a few small issues to iron out, I don't have a stable connection to discuss all of them in detail at the moment, but I don't think there is much left to do for the implementation.

huard:
Ok. I've added the Datacube extension to the STAC item. It lists variables and dimensions.

dchandan:
Ok, connection somewhat stable now, so pushing the new data model. Please check it out and give your feedback. Some other associated changes are in my Jupyter notebooks where I test things before transferring them to the actual source code. Will commit them when I am back from vac.

huard:
See https://github.com/crim-ca/stac-populator/pull/18 where I merged your work with mine. 

Essentially, I started to move not-cmip6 specific logic to stac_utils.py. I also fixed the PUT call, which needs the idem ID.

Open questions:
- Do we use pystac internally ? It offers some conveniences, but its inputs are not an exact match to the specs, which muddies things a bit. For example, start_datetime is an input to pystac.Item, but it then copies it to properties.
- How do we deal with non-conforming data. In examples I'm running, the `version` is missing. Is that a cause for rejecting the item ?

dchandan:
I do agree that pystac does some things in a weird way. pystac is only used right now in the `create_stac_collection` function. So it has a very limited use at the moment and can be removed and replaced with a pydantic model. I am fully onboard with doing that.

I think version is not a global attribute in the netcdf files CMORized for ESGF. If I recall correctly, I was planning to extract the version from the data path of the file (e.g. Root_location/CMIP6/CMIP/NCAR/CESM2/historical/r1i1p1f1/Amon/tas/gn/v20190308). All my data is structured in this CMIP6 hierarchy. I think version would be a necessary attribute because there can be differences in data files between versions.

implementations/CMIP6-UofT/add_CMIP6.py
```diff
@@ -34,19 +145,56 @@ def __init__(
...
+
+        a = STACAsset(
+            href=item_data["access_urls"]["HTTPServer"],
+            media_type="application/netcdf",
+            title="HTTP Server",
+            roles=["data"],
+        )
+
+        item = STACItem(
+            id="sdfdd",
```
Nazim-crim, line 183: Shouldn't this be an unique id ?

fmigneault:
@dchandan 
When can we expect this to be integrated?

I would like to synchronize with @huard's recent commits that resolve a lot of issues and pending https://github.com/crim-ca/stac-populator/pull/23 to combine everything and use it for https://github.com/crim-ca/ncml2stac/blob/main/notebooks/ncml2stac.ipynb and https://github.com/bird-house/birdhouse-deploy/pull/386. 

This is currently blocking us, but I don't want to push changes that would break what you are working on either.

dchandan:
Working on it. I'm changing things up a bit.

dchandan:
#23 is likely to be out of date after I push my refactoring.

fmigneault:
Ok. The liinks added in #23 will need to be transferred over because they are needed for https://github.com/bird-house/birdhouse-deploy/pull/386.

dchandan:
Closed because the PR was getting too long. Will fix outstanding issues in future PRs.

PR #17, opened 2023-08-28
Validation

huard:
Include CV validation when creating STAC item.


implementations/CMIP6-UofT/add_CMIP6.py
```diff
@@ -19,6 +22,116 @@
...
+    realization_index: int
+    forcing_index: int
+    variant_label: str
+    version: str
+    license: str = None
+    grid: str = None
+    tracking_id: str = Field(..., alias="tracking_id")
+
+
+def ncattrs_to_geometry(attrs: MutableMapping[str, Any]) -> MutableMapping[str, Any]:
```
dchandan, line 88: I wonder if we should make this a class method for `THREDDSLoader` so that they can be reused. The input keys these are expecting are very tied to the data returned by the THREDDS server.

implementations/CMIP6-UofT/add_CMIP6.py
```diff
@@ -50,15 +164,23 @@ def handle_ingestion_error(self, error: str, item_name: str, item_data: MutableM
...
-    def validate_stac_item_cv(self, data: MutableMapping[str, Any]) -> bool:
-        # TODO: next step is to implement this
-        pass
 
+        # Create STAC item geometry from CFMetadata
+        item = dict(
+            id = make_cmip6_id(item_data["attributes"]),
+            geometry = ncattrs_to_geometry(item_data["groups"]["CFMetadata"]["attributes"]),
+            bbox = ncattrs_to_bbox(item_data["groups"]["CFMetadata"]["attributes"]),
+            properties = Properties(item_data["attributes"]).dump_model(),
```
dchandan, line 173: This is very interesting. I haven't used Pydantic but I can see the convenience.

implementations/CMIP6-UofT/add_CMIP6.py
```diff
@@ -50,15 +164,23 @@ def handle_ingestion_error(self, error: str, item_name: str, item_data: MutableM
...
+            start_datetime = item_data["groups"]["CFMetadata"]["attributes"]["time_coverage_start"],
+            end_datetime = item_data["groups"]["CFMetadata"]["attributes"]["time_coverage_end"],
+        )
 
+        return pystac.Item(**item)
 
 
+    def validate_stac_item_cv(self, data: MutableMapping[str, Any]) -> bool:
```
dchandan, line 181: Yes, with the current method of validation, we can remove this.

implementations/CMIP6-UofT/add_CMIP6.py
huard: Makes sense, will do. Didn't want to integrate things to tightly until we got something working.

implementations/CMIP6-UofT/add_CMIP6.py
huard: It also allows us to rename the attributes (alias).

implementations/CMIP6-UofT/add_CMIP6.py
```diff
@@ -19,6 +22,116 @@
 LOGGER.propagate = False
 
 
+def collection2enum(collection):
```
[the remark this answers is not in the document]
huard: ok, will do.

PR #18, opened 2023-09-05
Replaced enums by literal for CMIP6 CV + Ingest items and assets

huard:


huard:
@dchandan

huard:
@Nazim-crim

Ça marche

STACpopulator/stac_utils.py
```diff
@@ -62,9 +70,100 @@ def collection2literal(collection):
...
+    method: PaginationMethods
+    body: Optional[Dict[Any, Any]] = None
+    merge: bool = False
+
+Links = RootModel[List[Union[PaginationLink, Link]]]
+
+
+class Item(BaseModel):
+    id: str = Field(..., alias="id", min_length=1)
+    geometry: Optional[Geometry] = None
```
dchandan, line 159: I don't think `geometry, bbox, properties` (maybe `assets` too?) should be optional. Not sure about `collection` and `datetime`. 

Also, I don't think we need `datetime` for the Item class since `start_datetime` and `end_datetime` are more appropriate for CMIP6 data and those are included in the `ItemProperties` data model.

STACpopulator/stac_utils.py
```diff
@@ -62,9 +70,100 @@ def collection2literal(collection):
...
+    bbox: Optional[List[float]] = None
+    properties: ItemProperties
+    assets: Dict[str, Asset] = None
+    stac_extensions: Optional[List[AnyUrl]] = []
+    collection: Optional[str] = None
+    datetime: Optional[dt.datetime] = None  # Not in the spec, but needed by pystac.Item.
+
 
 
 class CFJsonItem:
```
dchandan, line 169: I am note able to quite understand what this class is supposed to do....

Regarding conversion of a Pydantic data model, say `item` to a JSON that can be sent to the server with a POST request, I've found this to work just fine:

`j = json.loads(item.model_dump_json(by_alias=True))`

STACpopulator/input.py
dchandan, line 1: These changes are already included in #16

implementations/CMIP6-UofT/add_CMIP6.py
```diff
@@ -90,7 +91,8 @@ class STACItem(BaseModel):
 def make_cmip6_id(attrs: MutableMapping[str, Any]) -> str:
     """Return unique ID for CMIP6 data collection (multiple variables)."""
     keys = ["activity_id", "institution_id", "source_id", "experiment_id", "variant_label", "table_id", "grid_label",]
-    return "_".join(attrs[k] for k in keys)
+    item_name = "_".join(attrs[k] for k in keys)
```
dchandan: I've made a change to this function in #16. I included `variable_id` in the list so that the entire string becomes unique. I believe that string is what ESGF uses to uniquely identify each dataset. My thinking was that we could use the same string for the ID here. Is there an advantage to using a hash rather than a human readable string?

implementations/CMIP6-UofT/add_CMIP6.py
```diff
@@ -90,7 +91,8 @@ class STACItem(BaseModel):
...
+    return hashlib.md5(item_name.encode("utf-8")).hexdigest()
```
dchandan: I've made a change to this function in https://github.com/crim-ca/stac-populator/pull/16. I included variable_id in the list so that the entire string becomes unique. I believe that string is what ESGF uses to uniquely identify each dataset. My thinking was that we could use the same string for the ID here. Is there an advantage to using a hash rather than a human readable string?

STACpopulator/stac_utils.py
huard: The [specs](https://github.com/radiantearth/stac-spec/blob/master/item-spec/json-schema/item.json) mention that one of geometry or bbox is mandatory.

STACpopulator/input.py
huard: Not sure what the comment refers to exactly, but in the merge I moved some stuff around.

implementations/CMIP6-UofT/add_CMIP6.py
huard: One question I had was whether we wanted some form of variable aggregation at the item level. That is, I suspect one item could have different variables listed as assets, instead of having one item per variable. Not sure in practice what difference it would make, but I think this is worth testing from the client side. 

Unclear for me as well. I initially just copied the approach used for the collections. Playing with the clients, I realized that hashes for collections are really not practical. For items, I'm not sure they need to be human readable. I suspect the selection would be based on a search of properties, not parsing the item ID. So no strong opinion on this yet.

huard:
Oh crap. I forgot I had rights to push here now and I pushed directly to this branch instead of the Ouranos branch, hence merging my changes. Really sorry about this, that was not my intention.

PR #23, opened 2023-09-28
Added link to the THREDDS Catalog to the Collection object.

huard:


STACpopulator/stac_utils.py
```diff
@@ -190,6 +190,9 @@ class MySTACItem(Item):
         )
 
         item = pystac.Item(**json.loads(item.model_dump_json(by_alias=True)))
+        link = pystac.Link(rel="source", target=attrs["catalog_url"], media_type="text/html", title="thredds:" + attrs[
+            "id"])
```
fmigneault, line 194: Although I don't expect to have another `thredds` service on our instances, this `thredds:` value should be infered from the Magpie service-name before `/catalog/` part.

fmigneault (commented):
Do we want to include the following extension as well? 
- https://github.com/stac-extensions/cf
- https://github.com/stac-extensions/forecast

fmigneault:
@huard 
Looks good. 
Another issue I faced trying to port changes in https://github.com/crim-ca/ncml2stac is that https://github.com/Ouranosinc/stac-populator/tree/collection_link/implementations is not under the source directory, so it cannot be imported by other utilities.

fmigneault (approved):

tests/test_cmip6_extension.py
```diff
@@ -18,3 +18,30 @@ def test_extension():
...
+        if link["rel"] == "source":
+            assert link["href"] == thredds_catalog_URL
+            break
+
+    r = requests.get(stac_host + 'collections/CMIP6/items').json()
+    for link in r["features"][0]["links"]:
+        if link["rel"] == "source":
+            assert "thredds:" in link["title"]
+            break
+
```
fmigneault, line 47: Would need a `for.. else` clause to raise an error if the `break` case was not encountered.

STACpopulator/stac_utils.py
```diff
@@ -205,6 +206,18 @@ class MySTACItem(Item):
...
 
+    def item_link(self) -> pystac.Link:
+        """Return STAC Link to the catalog entry for the dataset."""
+        url = self.attrs["catalog_url"]
+        parts = url.split("/")
+        service = parts[parts.index("catalog") - 1]
+
+        link = pystac.Link(rel="source",
+                           target=url,
+                           media_type="text/html",
```
fmigneault, line 217: In the case of the STAC Item that represents a NetCDF, `source` should probably refer to `application/x-netcdf` with an URL of the actual file? This URL should be indicated at the top level of the NCML file as `@location` using `Dataset.to_cf_dict`.

If this change is done, instead of looking for `/catalog/` in the URL, the `/dodsC/` must be considered.

This code can be reused though, to inject `source` in the STAC Collection, which would be a XML THREDDS catalog URL.

fmigneault: I also found that the NCML has a `id` attribute in `THREDDSMetadata` with value `birdhouse/testdata/xclim/cmip6/sic_SImon_CCCma-CanESM5_ssp245_r13i1p2f1_2020.nc`. 
This seems more reliable than parsing the URL ourselves.

huard: 1. The href is not the URL of the actual netCDF file, but of its catalog entry, so not sure the mime should be application/netcdf. The netcdf link is listed among the Assets.
2.  Yes, I'm using that ID already.

tests/test_cmip6_extension.py
huard: Indeed, done.

STACpopulator/stac_utils.py
fmigneault: When loading in `xncml.Dataset`, the `"@location"` in returned attributes directly at the root is the NetCDF file URL. 

This is the same `location` attribute directly at the top of: 
https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/ncml/birdhouse/testdata/xclim/cmip6/sic_SImon_CCCma-CanESM5_ssp245_r13i1p2f1_2020.nc?catalog=https%3A%2F%2Fpavics.ouranos.ca%2Ftwitcher%2Fows%2Fproxy%2Fthredds%2Fcatalog%2Fbirdhouse%2Ftestdata%2Fxclim%2Fcmip6%2Fcatalog.html&dataset=birdhouse%2Ftestdata%2Fxclim%2Fcmip6%2Fsic_SImon_CCCma-CanESM5_ssp245_r13i1p2f1_2020.nc

The link actually returned uses `dodsC` service, so the response from THREDDS (https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/fileServer/birdhouse/testdata/xclim/cmip6/sic_SImon_CCCma-CanESM5_ssp245_r13i1p2f1_2020.nc) that I get is in fact a 400 HTML 😞 

What we could do is employ the `fileServer` endpoint instead, where that would be the real `source` in `application/x-netcdf` (https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/fileServer/birdhouse/testdata/xclim/cmip6/sic_SImon_CCCma-CanESM5_ssp245_r13i1p2f1_2020.nc).

I think the NCML XML href and the THREDDS Catalog HTML href would be relevant in `links` as `rel=describedby` and `rel=alternate` respectively. The NCML href could also use `convertedfrom` actually!

The `rel=collection` href could also be inserted since you have it.

STACpopulator/input.py
```diff
@@ -90,7 +104,8 @@ def extract_metadata(self, ds: siphon.catalog.Dataset) -> MutableMapping[str, An
         attrs = xncml.Dataset.from_text(r.content).to_cf_dict()
 
         attrs["access_urls"] = ds.access_urls
-
+        attrs["catalog_url"] = self.catalog_head.catalog_url
```
fmigneault, line 107: @huard 

For some reason, `ds.access_urls` causes an `AttributeError` on my side. Do you have an idea why that could be?

When I skip this step to avoid the error, my resulting STAC Item JSON (see bottom of output in https://github.com/crim-ca/ncml2stac/pull/3/files#diff-09442021b34927015b7a3703f6ed65611059a807c957feb4dfe2cd9955ccd68e) has the assets references, but the names are different: 
```json
  "assets": {
    "httpserver_service": { ... },
    "opendap_service": { ... },
    ...
  }
```

Seems to be caused by this logic: https://github.com/crim-ca/stac-populator/pull/23/files#diff-ea68b4da6eb5897aa6022852045603d912245daf3f0b044cb9abbd176e758455R196-R199

STACpopulator/stac_utils.py
fmigneault: https://github.com/crim-ca/ncml2stac/blob/main/notebooks/ncml2stac.ipynb contains the updated `item_link` implementation using `@location` attribute

fmigneault:
Should (?) be fixed via #25.

PR #28, opened 2023-11-08
add devtool utilities

fmigneault:
## Changes

* Add `LICENSE` file.
* Add `bump-my-version` with `make version` and `make VERSION=<...> bump` utilities to self-update release versions.
* Add more metadata to `pyproject.toml`.
* Adjust `README.md` with updated references and release version indicators.
* Add `CHANGES.md` to record version updates.
* Add `dev` dependencies to `pyproject.toml` for testing the package (install with `pip install ".[dev]"`).
* Remove `requirements.txt` in favor of all dependencies combined in `pyproject.toml`.
* Add test to validate STAC Collection and Item contain `source` with expected THREDDS format.
* Fix broken tests and invalid imports.

PR #30, opened 2023-11-09
Adding marble specific fields to item properties

dchandan:
This PR adds two properties to a STAC item, and in doing so helps describe aspects of the data or the STAC catalog that are useful on the Marble network: `mrbl:is_replica` and `mrbl:host_node`. 

`mrbl:host_node`, specifies the name of the Marble node on which the data is stored. Valid values for this (validity is not checked in this PR) are the values of the keys in the [Marble node registry](https://github.com/DACCS-Climate/Marble-node-registry/blob/current-registry/node_registry.json).

`mrbl:is_replica` is set to `false` in the STAC entry for an item when both the item and the catalog are on `mrbl:host_node`, and is `true` in any replicated STAC entries on any other node in the network (the purpose is very similar to the ESGF `replica` attribute, except on Marble the data is not replicated, only the catalog). 

The purpose of this PR is to get a sense of what others think about this, and to solicit useful feedback. "Implementation" itself is not critical for this PR, ideally, down the line, we would write a "Marble STAC extension" to add these properties and/or other future Marble related properties. Right now, we can just do it in a hack-ish way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ CATALOG = https://daccs.cs.toronto.edu/twitcher/ows/proxy/thredds/catalog/datase
 # CATALOG = https://daccs.cs.toronto.edu/twitcher/ows/proxy/thredds/catalog/datasets/CMIP6/CMIP/AS-RCEC/catalog.html
 
 testcmip6:
-	python $(IMP_DIR)/CMIP6_UofT/add_CMIP6.py $(STAC_HOST) $(CATALOG)
+	python $(IMP_DIR)/CMIP6_UofT/add_CMIP6.py $(STAC_HOST) $(CATALOG) PAVICS
 
 delcmip6:
 	curl --location --request DELETE '$(STAC_HOST)/collections/CMIP6_UofT'
--- a/STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
+++ b/STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
@@ -80,6 +80,8 @@ class CMIP6ItemProperties(STACItemProperties, validate_assignment=True):
     license: str
     grid: str
     mip_era: str
+    is_replica: bool = Field(..., serialization_alias="mrbl:is_replica")
+    host_node: str = Field(..., serialization_alias="mrbl:host_node")
 
     model_config = ConfigDict(alias_generator=add_cmip6_prefix, populate_by_name=True)
 
@@ -108,7 +110,9 @@ class CMIP6populator(STACpopulatorBase):
     item_properties_model = CMIP6ItemProperties
     item_geometry_model = GeoJSONPolygon
 
-    def __init__(self, stac_host: str, data_loader: GenericLoader, update: Optional[bool] = False) -> None:
+    def __init__(
+        self, stac_host: str, data_loader: GenericLoader, host_node: str, update: Optional[bool] = False
+    ) -> None:
         """Constructor
 
         :param stac_host: URL to the STAC API
@@ -117,6 +121,7 @@ class CMIP6populator(STACpopulatorBase):
         :type thredds_catalog_url: str
         """
         super().__init__(stac_host, data_loader, update)
+        self.host_node = host_node
 
     @staticmethod
     def make_cmip6_item_id(attrs: MutableMapping[str, Any]) -> str:
@@ -146,11 +151,20 @@ class CMIP6populator(STACpopulatorBase):
         """
         iid = self.make_cmip6_item_id(item_data["attributes"])
 
-        try:
-            item = STAC_item_from_metadata(iid, item_data, self.item_properties_model, self.item_geometry_model)
-        except pydantic_core._pydantic_core.ValidationError:
-            print(f"ERROR: ValidationError for {iid}")
-            return -1
+        # Adding some Marble network related data to the item properties.
+        # is_replica is always False when using this app since this app is only used to populate the hostnode
+        # with the data residing on that node. Replica STAC items on other nodes in the Marble network will be
+        # populated using a different app, and there the is_replica flag will be True.
+        item_data["attributes"]["is_replica"] = False
+        item_data["attributes"]["host_node"] = self.host_node
+
+        item = STAC_item_from_metadata(iid, item_data, self.item_properties_model, self.item_geometry_model)
+        # try:
+        #     item = STAC_item_from_metadata(iid, item_data, self.item_properties_model, self.item_geometry_model)
+        # except pydantic_core._pydantic_core.ValidationError as e:
+        #     print(f"ERROR: ValidationError for {iid}")
+        #     # e.errors()
+        #     return -1
 
         # Add the CMIP6 STAC extension
         item.stac_extensions.append(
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@ CATALOG = https://daccs.cs.toronto.edu/twitcher/ows/proxy/thredds/catalog/datase
 # CATALOG = https://daccs.cs.toronto.edu/twitcher/ows/proxy/thredds/catalog/datasets/CMIP6/CMIP/AS-RCEC/catalog.html
 
 testcmip6:
-	python $(IMP_DIR)/CMIP6_UofT/add_CMIP6.py $(STAC_HOST) $(CATALOG)
+	python $(IMP_DIR)/CMIP6_UofT/add_CMIP6.py $(STAC_HOST) $(CATALOG) PAVICS
```
fmigneault, line 9: I find this command is becoming too cluttered with optional arguments.
Can you add a `--host-node` option or similar?
I think it should use the URL of the `STAC_HOST` by default.

fmigneault (changes requested):
I would like to merge https://github.com/crim-ca/stac-populator/pull/28 first.
On follow PRs, tests will have to pass.
I add to patch some tests. Expected features/support were still not working.

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
```diff
@@ -146,11 +151,20 @@ def create_stac_item(self, item_name: str, item_data: MutableMapping[str, Any])
...
 
-        try:
-            item = STAC_item_from_metadata(iid, item_data, self.item_properties_model, self.item_geometry_model)
-        except pydantic_core._pydantic_core.ValidationError:
-            print(f"ERROR: ValidationError for {iid}")
-            return -1
+        # Adding some Marble network related data to the item properties.
+        # is_replica is always False when using this app since this app is only used to populate the hostnode
+        # with the data residing on that node. Replica STAC items on other nodes in the Marble network will be
+        # populated using a different app, and there the is_replica flag will be True.
```
huard, line 157: Not sure what is meant exactly by replica here. In the ESGF world, this has to do with who mirrors from whom. Will MARBLE include replication of datasets?

Also, more generally, if you are going to curate copies of ESGF files that you aggregate yourself, how are you going to determine whether one of the files making the aggregation has been updated, and this update needs to be propagated to your own version?

huard: Also, you say that replica items would be populated by another app. Why would that be? The replica is going to be a netCDF file sitting on a THREDDS server.

huard: Or do you mean the replica of a STAC item (just the catalog entry, not the file) ? In that case, I think we need to work out how STAC catalogs are going to be shared from one node to another.

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
```diff
@@ -80,6 +80,8 @@ class CMIP6ItemProperties(STACItemProperties, validate_assignment=True):
     license: str
     grid: str
     mip_era: str
+    is_replica: bool = Field(..., serialization_alias="mrbl:is_replica")
+    host_node: str = Field(..., serialization_alias="mrbl:host_node")
```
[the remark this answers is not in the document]
fmigneault: I think we could also avoid the `mrbl` shortcut, and simply have `marble` explicitly.
It is not really that much a long name to justify shortening it.

Makefile
dchandan: Yes, we can streamline this.

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
dchandan: > Not sure what is meant exactly by replica here. In the ESGF world, this has to do with who mirrors from whom. Will MARBLE include replication of datasets?

Not, datasets will not be replicated. We don't have enough space on the network. What I meant is that we have plans to ensure that STAC entries from one node are visible in the STAC browser of another node. As an example, at some point the Ouranos STAC catalog will also have a "CMIP6-UofT" collection whose entries be (near) duplicates of the entries on UofT's STAC catalog for the "CMIP6-UofT" (all the properties will be same and crucially, access links will be same), but `marble:host_node` property will be `UofT` and `marble:is_replica` property will be `true`. Similarly, UofT will have replica entries for catalogs on Ouranos. This way, a user can query any single node on the network to get info on data across the network.

> 
> Also, more generally, if you are going to curate copies of ESGF files that you aggregate yourself, how are you going to determine whether one of the files making the aggregation has been updated, and this update needs to be propagated to your own version?

It is my understanding that when files are updated on ESGF, they are marked with a new version number, existing files are never overridden. We should follow the same approach. The aggregations will be for all files in a single version. If, for example, one of the files in say an `EC-Earth` dataset (which come in 165 individual files) are updated, then it will be up to the manager of the node to decide if he/she wants to create a new aggregation using the updated file. My approach for UofT node would be to not touch do that unless the changes are substantial. Let me know if I didn't understand your question correctly.

dchandan: > Or do you mean the replica of a STAC item (just the catalog entry, not the file) ? In that case, I think we need to work out how STAC catalogs are going to be shared from one node to another.

Yes just STAC items. We will be doing this by completing the implementation of [STACLoader](https://github.com/crim-ca/stac-populator/blob/e7fd123c40ebd404ade128cd5c3fab7e6a167a34/STACpopulator/input.py#L134C1-L134C1). This will crawl a STAC catalog, get the JSON data, modify it as needed (for example to change `marble:is_replica`) and post it to another STAC catalog.

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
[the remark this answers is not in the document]
dchandan: Yeah, we should use `marble`.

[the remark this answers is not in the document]
dchandan: > These attributes would apply to every dataset hosted on the node, so I suggest they should be defined inside a Parent class that every collection inherits from.

I understood your intention and I think the proposal can be useful. But I didn't understand (or can't quite visualize) your proposal for how to do that. Could you please elaborate?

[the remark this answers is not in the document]
huard: This class inherits from `STACItemProperties`, which would be common to all implementations, correct ? So I'd add those two attributes in that class.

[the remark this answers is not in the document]
dchandan: Right I see.

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
fmigneault: Something to consider for mirrors or "replica": https://github.com/stac-extensions/alternate-assets

PR #33, opened 2023-11-15
Fix bugs in datacube extension. 

huard:
- Add test. 
- Add schema for local validation.


huard:
So the tests run fine locally, and I don't quite understand why the datacube one fails on the CI.

huard:
I traced the problem to pystac. Works with 1.8.3, fails with 1.9.0.

tests/test_cmip6_datacube.py
```diff
@@ -0,0 +1,33 @@
...
+    attrs["access_urls"] = {"HTTPServer": "http://example.com"}
+    item = STAC_item_from_metadata("test", attrs, CMIP6ItemProperties, GeoJSONPolygon)
+
+    # Add extension
+    dc = DataCubeHelper(attrs)
+    dc_ext = DatacubeExtension.ext(item, add_if_missing=True)
+    dc_ext.apply(dimensions=dc.dimensions, variables=dc.variables)
+
+    # Validate
+    schema_uri = DIR / "schemas" / "datacube.json"
```
dchandan, line 27: This is cool! Is this a new python syntax? Doesn't work with 3.10. If we want it to be compatible with different python versions then perhaps we should revert to a more traditional syntax.

pyproject.toml
```diff
@@ -78,6 +78,7 @@ dev = [
     "pytest-cov",
     "coverage",
     "bump-my-version",
+    "jsonschema",
```
dchandan, line 81: Maybe I missed something, but I don't see `jsonschema` being used anywhere. I think you are doing your validation with a function from `pystac`, is that right?

dchandan: Thanks for this @huard! 

About the version of the Datacube extension, how does a particular version of the extension get picked up? We never specify a version to use while applying the extension. Do different versions of `pystac` include different versions of the extension? When `add_CMIP6.py` runs on my machine, `pystac` applies version 2.0.0.

dchandan (approved):

pyproject.toml
dchandan: Okay, so to answer my own question, looks like different versions of `pystac` come with different versions of extensions. Specifically, for data cube, `pystac 1.8.3` comes with 2.0.0 and `pystac 1.9.0` comes with 2.2.0.

I think that in `test_cmip6_datacube.py`, for the validation schema, we should use the schema that is directly referenced in the `stac_extensions` property of the item being validated, rather than use a a [specific hard-coded version](https://github.com/crim-ca/stac-populator/pull/33/commits/93af1ac111e56070d87c7a241488724b8dfbaf68). This way, there would be no mis-match between the schema version used while creating the item (this directly comes from the version of `pystac` used) and the schema version used for validation.

tests/test_cmip6_datacube.py
huard: No, it's just syntactic sugar for the pathlib.Path object (see how DIR is defined). It's part of the standard lib and I find it clarifies manipulation of paths.

pyproject.toml
huard: I think it's required by pystac validation, but it may also be a left over from previous experimentation, not 100% sure. 

Yes, that's what I learned when trying to understand why the CI failed while it ran fine on my end. 

Yes, that makes a lot of sense.

dchandan:
Oh I see, this is from Pathlib.On Nov 16, 2023, at 9:00 AM, David Huard ***@***.***> wrote:﻿
@huard commented on this pull request.



In tests/test_cmip6_datacube.py:
> +
+
+def test_datacube_helper():
+    # Create item
+    ds = xncml.Dataset(filepath=DIR / "data" / "o3_Amon_GFDL-ESM4_historical_r1i1p1f1_gr1_185001-194912.xml")
+    attrs = ds.to_cf_dict()
+    attrs["access_urls"] = {"HTTPServer": "http://example.com"}
+    item = STAC_item_from_metadata("test", attrs, CMIP6ItemProperties, GeoJSONPolygon)
+
+    # Add extension
+    dc = DataCubeHelper(attrs)
+    dc_ext = DatacubeExtension.ext(item, add_if_missing=True)
+    dc_ext.apply(dimensions=dc.dimensions, variables=dc.variables)
+
+    # Validate
+    schema_uri = DIR / "schemas" / "datacube.json"

No, it's just syntactic sugar for the pathlib.Path object (see how DIR is defined). It's part of the standard lib and I find it clarifies manipulation of paths.

pyproject.toml
huard: The test now validates against the version used. If it's 2.2, it gets it from the repo, otherwise it will fetch it from the net. I find it useful when debugging to have a hard-copy of the schema nearby, but we could ditch it at some point.

huard:
I don't have permission to merge, let me know if there's anything else to modify.

dchandan:
I also can't seem to merge anymore. @fmigneault did you make any aches changes after taking over from Mathieu? Please give David and I elevated permissions.

fmigneault:
You should both be able to merge as long as there was 1 review approval from someone else.

tests/test_cmip6_datacube.py
```diff
@@ -0,0 +1,27 @@
...
+
+    # Add extension
+    dc = DataCubeHelper(attrs)
+    dc_ext = DatacubeExtension.ext(item, add_if_missing=True)
+    dc_ext.apply(dimensions=dc.dimensions, variables=dc.variables)
+
+    schemas = item.validate()
+    assert len(schemas) >= 2
+    assert "item.json" in schemas[0]
+    assert "datacube" in schemas[1]
```
fmigneault, line 27: Just to be sure, does that `datacube` refer to `"https://stac-extensions.github.io/datacube/v2.2.0/schema.json"` ?

tests/schemas/datacube/v2.2.0.json
fmigneault, line 1: Is the file used anywhere. I did not see a reference to it.

tests/test_cmip6_datacube.py
huard: Yes !

tests/schemas/datacube/v2.2.0.json
huard: It's handled by pystac I believe.

fmigneault: Tested locally. The schema does not seem to be needed. `pystac` automatically adds the reference from `pystac.extensions.datacube.SCHEMA_URI` and validates against `DatacubeExtension`.

PR #35, opened 2023-11-16
Add `ncattrs` function to get nc attributes without the need for siphon.

huard:
I didn't modify THREDDSLoader yet. Just added the function and tests. 

@fmigneault-crim I replaced the comparison with the reference with validation. The reference was missing a few services, so the test failed.

### DIFF
--- a/STACpopulator/stac_utils.py
+++ b/STACpopulator/stac_utils.py
@@ -79,6 +79,56 @@ def collection2literal(collection, property="label"):
     return Literal[terms]
 
 
+def thredds_catalog_attrs(url: str) -> dict:
+    """Return attributes from the catalog.xml THREDDS server response."""
+    import xmltodict
+    import requests
+
+    xml = requests.get(url).text
+
+    raw = xmltodict.parse(
+        xml,
+        process_namespaces=True,
+        namespaces={
+            "http://www.unidata.ucar.edu/namespaces/thredds/InvCatalog/v1.0": None,
+            "https://www.unidata.ucar.edu/namespaces/thredds/InvCatalog/v1.0": None,
+        },
+    )
+    return raw
+
+
+def ncattrs(url: str) -> dict:
+    """Return attributes from a THREDDS netCDF dataset."""
+    import requests
+    import xncml
+    import urllib
+
+    pr = urllib.parse.urlparse(url)
+
+    parts = url.split("/")
+    nc = parts[-1]
+
+    # Get catalog information about available services
+    catalog = "/".join(parts[:-1]) + "/catalog.xml"
+    cattrs = thredds_catalog_attrs(catalog)["catalog"]
+
+    cid = cattrs["dataset"]["@ID"]
+
+    # Get service URLs for the dataset
+    access_urls = {}
+    for service in cattrs["service"]["service"]:
+        access_urls[service["@serviceType"]] = f'{pr.scheme}://{pr.netloc}{service["@base"]}{cid}/{nc}'
+
+    # Get dataset attributes
+    r = requests.get(access_urls["NCML"])
+    attrs = xncml.Dataset.from_text(r.text).to_cf_dict()
+    attrs["attributes"] = numpy_to_python_datatypes(attrs["attributes"])
+
+    # Include service attributes
+    attrs["access_urls"] = access_urls
+    return attrs
+
+
 def ncattrs_to_geometry(attrs: MutableMapping[str, Any]) -> MutableMapping[str, Any]:
     """Create Polygon geometry from CFMetadata."""
     attrs = attrs["groups"]["CFMetadata"]["attributes"]
--- a/tests/test_standalone_stac_item.py
+++ b/tests/test_standalone_stac_item.py
@@ -1,16 +1,15 @@
 import json
 import pytest
-import requests
 import os
 import tempfile
 from urllib.parse import quote
 
-import xncml
-
 from STACpopulator.implementations.CMIP6_UofT.add_CMIP6 import CMIP6ItemProperties, CMIP6populator
 from STACpopulator.input import THREDDSLoader
 from STACpopulator.models import GeoJSONPolygon
-from STACpopulator.stac_utils import STAC_item_from_metadata
+from STACpopulator.stac_utils import STAC_item_from_metadata, ncattrs
+from pystac.validation import JsonSchemaSTACValidator
+from pystac import STACObjectType
 
 CUR_DIR = os.path.dirname(__file__)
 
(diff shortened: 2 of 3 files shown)


STACpopulator/stac_utils.py
```diff
@@ -79,6 +79,56 @@ def collection2literal(collection, property="label"):
     return Literal[terms]
 
 
+def thredds_catalog_attrs(url: str) -> dict:
+    """Return attributes from the catalog.xml THREDDS server response."""
+    import xmltodict
+    import requests
```
fmigneault, line 85: Move imports at the top of the file.

STACpopulator/stac_utils.py
```diff
@@ -79,6 +79,56 @@ def collection2literal(collection, property="label"):
...
+        },
+    )
+    return raw
+
+
+def ncattrs(url: str) -> dict:
+    """Return attributes from a THREDDS netCDF dataset."""
+    import requests
+    import xncml
+    import urllib
```
fmigneault, line 104: idem

tests/test_standalone_stac_item.py
```diff
@@ -21,35 +20,11 @@ def quote_none_safe(url):
...
-    attrs = xncml.Dataset.from_text(data).to_cf_dict()
-    attrs["access_urls"] = {  # FIXME: all following should be automatically added, but they are not!
-        "HTTPServer": f"{thredds_url}/fileServer/{thredds_path}/{thredds_nc}",
-        "OPENDAP": f"{thredds_url}/dodsC/{thredds_path}/{thredds_nc}",
-        "WCS": f"{thredds_url}/wcs/{thredds_path}/{thredds_nc}?service=WCS&version=1.0.0&request=GetCapabilities",
-        "WMS": f"{thredds_url}/wms/{thredds_path}/{thredds_nc}?service=WMS&version=1.3.0&request=GetCapabilities",
-        "NetcdfSubset": f"{thredds_url}/ncss/{thredds_path}/{thredds_nc}/dataset.html",
-    }
-
+    url = "https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/birdhouse/testdata/xclim/cmip6/sic_SImon_CCCma-CanESM5_ssp245_r13i1p2f1_2020.nc"
```
fmigneault, line 23: I suggest using `@pytest.mark.parametrize()` with the `url` as input. 

We could test the PAVICS URL as well as the https://psl.noaa.gov/thredds/catalog/Datasets/catalog.html that @dchandan referred to as problematic.

tests/test_standalone_stac_item.py
```diff
@@ -21,35 +20,11 @@ def quote_none_safe(url):
...
+    attrs = ncattrs(url)
     stac_item_id = CMIP6populator.make_cmip6_item_id(attrs["attributes"])
     stac_item = STAC_item_from_metadata(stac_item_id, attrs, CMIP6ItemProperties, GeoJSONPolygon)
-
-    ref_file = os.path.join(CUR_DIR, "data/stac_item_testdata_xclim_cmip6_ncml.json")
-    with open(ref_file, mode="r", encoding="utf-8") as ff:
-        reference = json.load(ff)
-
-    assert stac_item.to_dict() == reference
+    assert stac_item.validate()
```
fmigneault, line 27: I think that might not be validating all the contents, only the CMIP6 schema. 

That would be because `CMIP6ItemProperties` is based on `pydantic` schema, while `pystac.Item` use their own `pystac.validation.STACValidator`.

tests/test_standalone_stac_item.py
```diff
@@ -21,35 +20,11 @@ def quote_none_safe(url):
...
+    url = "https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/birdhouse/testdata/xclim/cmip6/sic_SImon_CCCma-CanESM5_ssp245_r13i1p2f1_2020.nc"
+    attrs = ncattrs(url)
     stac_item_id = CMIP6populator.make_cmip6_item_id(attrs["attributes"])
     stac_item = STAC_item_from_metadata(stac_item_id, attrs, CMIP6ItemProperties, GeoJSONPolygon)
-
-    ref_file = os.path.join(CUR_DIR, "data/stac_item_testdata_xclim_cmip6_ncml.json")
-    with open(ref_file, mode="r", encoding="utf-8") as ff:
-        reference = json.load(ff)
-
-    assert stac_item.to_dict() == reference
```
fmigneault, line 52: I think we should keep the explicit validation with expected contents. This way, we make sure that the methods did not impact other parts of the properties/metadata shuffling.

fmigneault (changes requested):
The new function is not employed by the populator?

tests/test_standalone_stac_item.py
huard: parametrize: ok

But then we need a reference for a link that might change at any time, no ?

tests/test_standalone_stac_item.py
huard: In theory, `validate()` returns the schemas it used to validate the content. Here it returns the STAC item schema. 
It validates only the STAC item for now (no extensions).

tests/test_standalone_stac_item.py
huard: But it makes it a bit more difficult to test a wider variety of files, i.e. we'll need a reference for each one of those, and we'll have to keep our hard-coded tests in sync with the version found on live servers.

tests/test_standalone_stac_item.py
huard: validate calls the pystac validator based on the JSONSchema. The test is more explicit about this now.

tests/test_standalone_stac_item.py
fmigneault: Good point.
Instead, we could make a local copy of the XML content and leave a comment of the reference URL in case we need to look it up later.
Could be the same for the PAVICS URL as well.

tests/test_standalone_stac_item.py
fmigneault: Function `STAC_item_from_metadata` returns a `STACpopulator.stac_utils.STACItem` derived from `pydantic.BaseModel`.

Unless it does something special, I don't think that would call the `pystac` validator. It would if the object returned was `pystac.Item`. 

Note, I'm trying to see if we can get rid of this dual validator approach in https://github.com/crim-ca/stac-populator/pull/32

tests/test_standalone_stac_item.py
fmigneault: I think we should have a set of `{test-ncml-file}.xml` / `{expected-stac-item}.json`, as simply load corresponding paths with `@pytest.mark.parametrize`. 

With local references, things shouldn't drastically change unless we ourselves modify the implementation.

tests/test_standalone_stac_item.py
huard: Argh, my bad. I thought we had a pystac.Item object...

huard: In fact we do !  `STAC_item_from_metadata` returns a pystac.Item:
```
```

tests/test_standalone_stac_item.py
huard: It's going to be a bit messier, since we need the catalog.xml and the netCDF xml, but yes, I can work on that.

huard: Another problem is that the catalog.xml does not include the host name. So the logic to reconstruct access_urls would have to be different when reading from an xml file on disk and when making a server request, which defeats the purpose of unit tests.

huard: The @serviceType varies across servers (e.g. OPENDAP, OpenDAP). Maybe better to use the @name (odap).

fmigneault: > Another problem is that the catalog.xml does not include the host name. So the logic to reconstruct access_urls would have to be different when reading from an xml file on disk and when making a server request, which defeats the purpose of unit tests.

I think instead you can use `responses` to mock requests.
So you pass in the URL as if it was the real instance, but with swaps the response with the XML on disk. 

> The `@serviceType` varies across servers (e.g. OPENDAP, OpenDAP). Maybe better to use the `@name` (odap).

I think we should use the standard names (in schema `ServiceTypeEnum`).
However, we might need some kind of custom handling because even the docs seems to provide multiple casing variations: 
- https://docs.unidata.ucar.edu/tds/current/userguide/services_ref.html#standard-data-services
- https://docs.unidata.ucar.edu/tds/current/userguide/services_ref.html#tds-requirements-for-thredds-catalog-service-elements
- https://docs.unidata.ucar.edu/tds/current/userguide/tds_config_ref.html#opendap-service
- https://schemas.unidata.ucar.edu/thredds/InvCatalog.1.2.xsd (see `ServiceTypeEnum`)

I see at least `OPENDAP`, `OPeNDAP`, `opendap`, `Opendap`, and `OpenDAP`.

tests/test_standalone_stac_item.py
huard: Done using responses.

tests/test_standalone_stac_item.py
huard: conftest.py now has a function to record the server's response. It's stored in the test directory. 
Another function used the stored response to create STAC item dicts, which can then be used by unit tests to compare results from live servers with a reference.

Added Enum with the serviceTypes defined in the latest version of the schema (1.0.7).

fmigneault (approved):

fmigneault:
@huard 
I can't fix the conflict myself since the branch is within Ouranos repo.
Good to merge after this is fixed.
Then, I can release a new version.

huard:
Maybe releasing this should wait for the functions to be integrated in the populator. As it it, they're just there.

fmigneault:
> Maybe releasing this should wait for the functions to be integrated in the populator.

Agreed. 
Let us know when this is ready after integration into the populator.

PR #44, opened 2024-01-17
Fixing a bug in thredds loader that limited crawling ability

dchandan:
A subtle bug was limiting the number of files that could be successfully crawled by the crawler before the crawler abruptly stopped (without error). Investigation released that this had to do with the "depth" variable in the `THREDDSLoader` class, that was [originally](https://github.com/crim-ca/stac-ingest/blob/e4cc2a66fee4b86ec238f139135d78215ec91ea4/stac_ingest/utils/tds.py) put in place to limit the depth to which the crawler will crawl. This original logic was largely retained, however, it turns out, there was an error in the logic which only showed up while crawling several deeply-nested data structures like the datasets in the [UofT CMIP6 data catalog](https://daccs.cs.toronto.edu/twitcher/ows/proxy/thredds/catalog/datasets/CMIP6/catalog.html). Essentially, the depth variable was always linearly decreasing and when it became 0, the crawler stopped working. This PR fixes that issue.

### DIFF
--- a/STACpopulator/input.py
+++ b/STACpopulator/input.py
@@ -60,6 +60,7 @@ class THREDDSCatalog(TDSCatalog):
     Because of how :class:`TDSCatalog` automatically loads and parses right away from ``__init__`` call,
     we need to hack around how the ``session`` attribute gets defined.
     """
+
     def __init__(self, catalog_url: str, session: Optional[Session] = None) -> None:
         self._session = session
         super().__init__(catalog_url)
@@ -91,7 +92,8 @@ class THREDDSLoader(GenericLoader):
         :type depth: int, optional
         """
         super().__init__()
-        self._depth = depth if depth is not None else 1000
+        self._max_depth = depth if depth is not None else 1000
+        self._depth = 0
 
         self.thredds_catalog_URL = self.validate_catalog_url(thredds_catalog_url)
 
@@ -134,18 +136,23 @@ class THREDDSLoader(GenericLoader):
         """Reset the generator."""
         self.catalog_head = self.catalog
 
-    def __iter__(self) -> Iterator[Tuple[str, MutableMapping[str, Any]]]:
+    def __iter__(self) -> Iterator[Tuple[str, str, MutableMapping[str, Any]]]:
         """Return a generator walking a THREDDS data catalog for datasets."""
+
+        if self._depth > self._max_depth:
+            return
+
         if self.catalog_head.datasets.items():
             for item_name, ds in self.catalog_head.datasets.items():
                 attrs = self.extract_metadata(ds)
-                yield item_name, attrs
+                yield item_name, ds.url_path, attrs
+                # yield item_name, ds.url_path, []
 
-        if self._depth > 0:
-            for name, ref in self.catalog_head.catalog_refs.items():
-                self.catalog_head = ref.follow()
-                self._depth -= 1
-                yield from self
+        for name, ref in self.catalog_head.catalog_refs.items():
+            self.catalog_head = ref.follow()
+            self._depth -= 1
+            yield from self
+            self._depth += 1
 
     def __getitem__(self, dataset):
         return self.catalog.datasets[dataset]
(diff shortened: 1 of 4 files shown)


STACpopulator/input.py
```diff
@@ -201,7 +207,7 @@ def __iter__(self) -> Iterator[Tuple[str, MutableMapping[str, Any]]]:
                 if self.prune:  # stop recursive search if requested
                     del dirs[:]
                 col_path = os.path.join(root, self._collection_name)
-                yield col_path, self._load_json(col_path)
+                yield col_path, "", self._load_json(col_path)
```
fmigneault, line 210: Since item location is now returned, this should be updated properly to do 
`yield self._collection_name, self._load_json(col_path)`

STACpopulator/input.py
```diff
@@ -211,7 +217,7 @@ def __iter__(self) -> Iterator[Tuple[str, MutableMapping[str, Any]]]:
             for name in files:
                 if not self._collection_mode and self._is_item(name):
                     item_path = os.path.join(root, name)
-                    yield item_path, self._load_json(item_path)
+                    yield item_path, "", self._load_json(item_path)
```
fmigneault, line 220: Same as other comment.

STACpopulator/populator_base.py
```diff
@@ -144,9 +143,12 @@ def publish_stac_collection(self, collection_data: dict[str, Any]) -> None:
...
 
     def ingest(self) -> None:
+        counter = 0
         LOGGER.info("Data ingestion")
-        for item_name, item_data in self._ingest_pipeline:
-            LOGGER.info(f"Creating STAC representation for {item_name}")
+        for item_name, item_loc, item_data in self._ingest_pipeline:
+            LOGGER.info(f"New data item: {item_name}")
+            if item_loc:
+                LOGGER.info(f"Data location: {item_loc}")
```
fmigneault, line 151: The `if` can be removed once the paths are returned for directory loader.

STACpopulator/populator_base.py
```diff
@@ -157,3 +159,5 @@ def ingest(self) -> None:
                     update=self.update,
                     session=self._session,
                 )
+                counter += 1
+                LOGGER.info(f"Processed {counter} data items")
```
fmigneault, line 163: Should there be a warning log if the STAC item was not generated/returned? Possibility an unexpected error that was silently ignored?

STACpopulator/input.py
```diff
@@ -134,18 +136,22 @@ def reset(self):
         """Reset the generator."""
         self.catalog_head = self.catalog
 
-    def __iter__(self) -> Iterator[Tuple[str, MutableMapping[str, Any]]]:
+    def __iter__(self) -> Iterator[Tuple[str, str, MutableMapping[str, Any]]]:
         """Return a generator walking a THREDDS data catalog for datasets."""
```
fmigneault, line 140: The yield return in the docstring would be helpful to indicate what is expected for each of the tuple elements.
Otherwise, we have to look somewhere else like inside `ingest()` to figure out what it should be.

STACpopulator/input.py
dchandan: Done.

STACpopulator/populator_base.py
dchandan: Done.

STACpopulator/populator_base.py
dchandan: Yes, good point. I've been wanting to add some more error management code. For the moment, I have added an error log message here.

STACpopulator/input.py
dchandan: Done.

huard (approved):
Good catch, sorry I missed this.

STACpopulator/populator_base.py
fmigneault: Can you add the item name and loc to the log error message so we can debug the one that failed appropriately?

fmigneault (commented):
Just last logging comment. Then, good to merge.

dchandan:
> Good catch, sorry I missed this.

I missed it too!

dchandan:
> Just last logging comment. Then, good to merge.

I've added this to #45.

PR #45, opened 2024-01-22
Adding feature to intercept errors and save information to files

dchandan:
Adding feature to intercept errors encountered while (i) generating STAC items, (ii) posting STAC items to the server. Information about the dataset and the error it encountered are saved to separate log files. An example log file from a run of the stac-populator is included below.

[stac-populator_CMIP6populator_errors_20240119-185007.txt](https://github.com/crim-ca/stac-populator/files/14012790/stac-populator_CMIP6populator_errors_20240119-185007.txt)

### DIFF
--- a/STACpopulator/api_requests.py
+++ b/STACpopulator/api_requests.py
@@ -73,7 +73,7 @@ def post_stac_item(
     json_data: dict[str, dict],
     update: Optional[bool] = True,
     session: Optional[Session] = None,
-) -> Union[None, str]:
+) -> None:
     """Post a STAC item to the host server.
 
     :param stac_host: address of the STAC host
@@ -99,12 +99,10 @@ def post_stac_item(
         if update:
             LOGGER.info(f"Item {item_id} already exists. Updating.")
             r = session.put(os.path.join(stac_host, f"collections/{collection_id}/items/{item_id}"), json=json_data)
-            return f"Requests: {r.reason}"
-            # r.raise_for_status()
+            # return f"Requests: {r.reason}"
+            r.raise_for_status()
         else:
             LOGGER.warn(f"Item {item_id} already exists.")
     else:
-        return f"Requests: {r.reason}"
-        # r.raise_for_status()
-
-    return None
+        # return f"Requests: {r.reason}"
+        r.raise_for_status()
--- a/STACpopulator/api_requests.py
+++ b/STACpopulator/api_requests.py
@@ -99,10 +99,8 @@ def post_stac_item(
         if update:
             LOGGER.info(f"Item {item_id} already exists. Updating.")
             r = session.put(os.path.join(stac_host, f"collections/{collection_id}/items/{item_id}"), json=json_data)
-            # return f"Requests: {r.reason}"
             r.raise_for_status()
         else:
             LOGGER.warn(f"Item {item_id} already exists.")
     else:
-        # return f"Requests: {r.reason}"
         r.raise_for_status()
--- a/STACpopulator/api_requests.py
+++ b/STACpopulator/api_requests.py
@@ -1,10 +1,10 @@
 import logging
 import os
-from typing import Any, Optional
+from typing import Any, Optional, Union
 
 import requests
-from requests import Session
 from colorlog import ColoredFormatter
+from requests import Session
 
 LOGGER = logging.getLogger(__name__)
 LOGFORMAT = "  %(log_color)s%(levelname)s:%(reset)s %(blue)s[%(name)-30s]%(reset)s %(message)s"
@@ -81,7 +81,7 @@ def post_stac_item(
     json_data: dict[str, dict],
     update: Optional[bool] = True,
     session: Optional[Session] = None,
-) -> None:
+) -> Union[None, str]:
     """Post a STAC item to the host server.
 
     :param stac_host: address of the STAC host
@@ -107,8 +107,12 @@ def post_stac_item(
         if update:
             LOGGER.info(f"Item {item_id} already exists. Updating.")
             r = session.put(os.path.join(stac_host, f"collections/{collection_id}/items/{item_id}"), json=json_data)
-            r.raise_for_status()
+            return f"Requests: {r.reason}"
+            # r.raise_for_status()
         else:
-            LOGGER.info(f"Item {item_id} already exists.")
+            LOGGER.warn(f"Item {item_id} already exists.")
     else:
-        r.raise_for_status()
+        return f"Requests: {r.reason}"
+        # r.raise_for_status()
+
+    return None
--- a/STACpopulator/api_requests.py
+++ b/STACpopulator/api_requests.py
@@ -3,17 +3,9 @@ import os
 from typing import Any, Optional, Union
 
 import requests
-from colorlog import ColoredFormatter
 from requests import Session
 
 LOGGER = logging.getLogger(__name__)
-LOGFORMAT = "  %(log_color)s%(levelname)s:%(reset)s %(blue)s[%(name)-30s]%(reset)s %(message)s"
-formatter = ColoredFormatter(LOGFORMAT)
-stream = logging.StreamHandler()
-stream.setFormatter(formatter)
-LOGGER.addHandler(stream)
-LOGGER.setLevel(logging.INFO)
-LOGGER.propagate = False
 
 
 def stac_host_reachable(url: str, session: Optional[Session] = None) -> bool:
--- a/STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
+++ b/STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
@@ -1,5 +1,6 @@
 import argparse
 import json
+import logging
 import os
 from typing import Any, MutableMapping, NoReturn, Optional, Union
 
@@ -13,9 +14,8 @@ from STACpopulator.extensions.thredds import THREDDSExtension, THREDDSHelper
 from STACpopulator.input import ErrorLoader, GenericLoader, THREDDSLoader
 from STACpopulator.models import GeoJSONPolygon
 from STACpopulator.populator_base import STACpopulatorBase
-from STACpopulator.stac_utils import get_logger
 
-LOGGER = get_logger(__name__)
+LOGGER = logging.getLogger(__name__)
 
 
 class CMIP6populator(STACpopulatorBase):
--- a/STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
+++ b/STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
@@ -29,6 +29,7 @@ class CMIP6populator(STACpopulatorBase):
         update: Optional[bool] = False,
         session: Optional[Session] = None,
         config_file: Optional[Union[os.PathLike[str], str]] = None,
+        log_debug: Optional[bool] = False,
     ) -> None:
         """Constructor
 
@@ -37,11 +38,7 @@ class CMIP6populator(STACpopulatorBase):
         :param data_loader: loader to iterate over ingestion data.
         """
         super().__init__(
-            stac_host,
-            data_loader,
-            update=update,
-            session=session,
-            config_file=config_file,
+            stac_host, data_loader, update=update, session=session, config_file=config_file, log_debug=log_debug
         )
 
     def create_stac_item(
@@ -101,6 +98,7 @@ def make_parser() -> argparse.ArgumentParser:
             "By default, uses the adjacent configuration to the implementation class."
         ),
     )
+    parser.add_argument("--debug", action="store_true", help="Set logger level to debug")
     add_request_options(parser)
     return parser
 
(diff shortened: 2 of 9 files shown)


huard:
I think this PR branches off the other one about the crawler. I'll wait for you to merge the first one before reviewing.

dchandan:
I've merged it now.

huard (commented):
Maybe dumb question, but why are we using an "error file" in the first place and not just using logger to log errors to file ?

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
```diff
@@ -69,15 +71,15 @@ def create_stac_item(self, item_name: str, item_data: MutableMapping[str, Any])
...
+            return (-1, "Failed to add Datacube extension")
 
         try:
             thredds_helper = THREDDSHelper(item_data["access_urls"])
             thredds_ext = THREDDSExtension.ext(item)
             thredds_ext.apply(thredds_helper.services, thredds_helper.links)
         except Exception:
             LOGGER.error("Failed to add THREDDS references to item %s", item_name)
-            raise
+            return (-1, "Failed to add THREDDS references")
```
fmigneault, line 82: This and other above should instead make use of pythonic `raise <StacPopulatorException>("message")`. 
The parent operation then decides what to do with the exception.

STACpopulator/api_requests.py
```diff
@@ -107,8 +107,12 @@ def post_stac_item(
...
-            r.raise_for_status()
+            return f"Requests: {r.reason}"
+            # r.raise_for_status()
         else:
-            LOGGER.info(f"Item {item_id} already exists.")
+            LOGGER.warn(f"Item {item_id} already exists.")
     else:
-        r.raise_for_status()
+        return f"Requests: {r.reason}"
+        # r.raise_for_status()
```
fmigneault, line 116: I think it is more appropriate to raise the exception and let the parent code handle it accordingly.
For example, a 403 would make all operations useless, while 409 could be ok if override was not requested, and 502 could indicate that the endpoint is wrong entirely. 
This information is loss with this strategy.

huard:
It's possible to configure the LOGGER object to write both to the terminal and to a file on disk. So if that meets the requirements, it could be an option. 

In another project I've used this:
```
def init_log():
    logger = logging.getLogger(__name__)
    logger.setLevel(logging.ERROR)

    fh = logging.FileHandler('log/run_MAGICC.log')
    fh.setLevel(logging.INFO)
    logger.addHandler(fh)
    return logger 
```
which logs ERRORs to the terminal, but stores all INFO messages to a file.

STACpopulator/populator_base.py
```diff
@@ -142,23 +142,39 @@ def create_stac_collection(self) -> dict[str, Any]:
     def publish_stac_collection(self, collection_data: dict[str, Any]) -> None:
         post_stac_collection(self.stac_host, collection_data, self.update, session=self._session)
 
+    def open_error_file(self) -> TextIO:
+        implementation_name = type(self).__name__
+        fname = f"stac-populator_{implementation_name}_errors_{datetime.strftime(datetime.now(), '%Y%m%d-%H%M%S')}.txt"
```
fmigneault, line 147: This looks a lot like a log formatter format. One should be used.

STACpopulator/populator_base.py
```diff
@@ -142,23 +142,39 @@ def create_stac_collection(self) -> dict[str, Any]:
     def publish_stac_collection(self, collection_data: dict[str, Any]) -> None:
         post_stac_collection(self.stac_host, collection_data, self.update, session=self._session)
 
+    def open_error_file(self) -> TextIO:
```
fmigneault, line 145: This is essentially what `logging.FileHandler` does, but it properly handles thread-safe operations as well.

fmigneault (changes requested):
Like @huard mentioned, I think it would work better if the logger was reused. 
One way to implement it would be to add multiple log handlers. 
This way, you only need to call `log.<level>()` methods once, and each handler performs whatever it needs to do with the record respectively.

```python 
logger = get_logger("stac-populator")
log_handler = logging.StreamHandler(sys.stderr)
log_handler.setFormatter(logging._defaultFormatter)  # or whatever
err_handler = logging.FileHandler("<file-path>")
err_handler.setFormatter(logging.Formatter(fmt="{stac_item_name}: {stac_item_loc}"))
logger.addHandler(log_handler)
logger.addHandler(err_handler)

logger.debug("Custom message: {n}: {l}", n=stac_item_name, l=stac_item_loc)
```

A custom formatter could be defined to ignore the records when required arguments (stac_item_name, stac_item_loc, etc.) are missing. Just need to define the `emit` method that handles the log record.

fmigneault:
@dchandan 
A timely video that came up in my watch list today: https://www.youtube.com/watch?v=9L77QExPmI0 
See JSON logs and extra parameter timestamps for some inspiration.

dchandan:
Thanks for the comments and the useful video on the log file. I do like the JSON lines output format shown in the video as it would make parsing the log files for issues much easier. I've now revamped the whole app logging to use the setup in the video.

STACpopulator/populator_base.py
```diff
@@ -142,23 +146,39 @@ def create_stac_collection(self) -> dict[str, Any]:
     def publish_stac_collection(self, collection_data: dict[str, Any]) -> None:
         post_stac_collection(self.stac_host, collection_data, self.update, session=self._session)
 
+    def configure_app_logging(self, log_debug) -> None:
+        """Configure the logger for the App."""
+        # generating the log file name
+        implementation_name = type(self).__name__
+        fname = f"{implementation_name}_log_{datetime.strftime(datetime.now(), '%Y%m%d-%H%M%S')}.jsonl"
+        log_level = "DEBUG" if log_debug else "INFO"
+        setup_logging(fname, log_level)
```
fmigneault, line 155: I suggest using `logging.INFO`/`logging.DEBUG` and use `logging.getLevelName` in the function for the configuration.

STACpopulator/populator_base.py
```diff
@@ -142,23 +146,39 @@ def create_stac_collection(self) -> dict[str, Any]:
     def publish_stac_collection(self, collection_data: dict[str, Any]) -> None:
         post_stac_collection(self.stac_host, collection_data, self.update, session=self._session)
 
+    def configure_app_logging(self, log_debug) -> None:
```
fmigneault, line 149: missing typing for `log_debug`

STACpopulator/populator_base.py
```diff
@@ -142,23 +146,39 @@ def create_stac_collection(self) -> dict[str, Any]:
...
+        """Configure the logger for the App."""
+        # generating the log file name
+        implementation_name = type(self).__name__
+        fname = f"{implementation_name}_log_{datetime.strftime(datetime.now(), '%Y%m%d-%H%M%S')}.jsonl"
```
fmigneault, line 153: Consider `utcnow` with datetime format using `Z` or UTC timezone notation ?

STACpopulator/populator_base.py
```diff
@@ -39,6 +42,7 @@ def __init__(
         """
 
         super().__init__()
+        self.configure_app_logging(log_debug)
```
fmigneault, line 45: Seems odd to me that logging would be configured here, especially since logging can be performed by other modules in the package. 
I think it should be done in the CLI where debug/log-level could be configured/modified by options: https://github.com/crim-ca/stac-populator/blob/10b07161b5690c82d88270bd39ad73e54d02340a/STACpopulator/cli.py#L159

Example of a set of options setup for the CLI: 
https://github.com/crim-ca/weaver/blob/d911c161721c2e25ee4663d34a49d2b218261710/weaver/cli.py#L1785-L1804

STACpopulator/logging.py
```diff
@@ -0,0 +1,128 @@
...
+    "version": 1,
+    "disable_existing_loggers": False,
+    "formatters": {
+        "simple": {
+            "()": "colorlog.ColoredFormatter",
+            "format": "  %(log_color)s%(levelname)s:%(reset)s %(blue)s[%(name)-30s]%(reset)s %(message)s",
+            "datefmt": "%Y-%m-%dT%H:%M:%S%z",
+        },
+        "json": {
+            "()": "STACpopulator.logging.JSONLogFormatter",
```
fmigneault, line 100: Consider `f"{JSONLogFormatter.__module__}.{JSONLogFormatter.__name__}"` so that IDEs can resolve references and make the config dynamic with the actual location.

STACpopulator/populator_base.py
dchandan: Sure... 🤷‍♂️ Don't see any obvious or materially significant benefit from this..

STACpopulator/populator_base.py
dchandan: You mean something like `datetime.strftime(datetime.utcnow(), '%Y%m%dT%H%M%S')` ?

STACpopulator/populator_base.py
dchandan: Doing it in the populator_base allows me to extract the name of the specific derived implementation that is being used, which I then use to name the log file.

fmigneault: The file name could be found the same way since the implementation to use is one of the required parameters of the CLI.

STACpopulator/populator_base.py
fmigneault: `datetime.utcnow().isoformat() + "Z"` or add the `datetime.timezone` object to the date returned by `datetime.now()`, and then `.isoformat()` that TZ-aware datetime to get something like `"2024-02-20T01:22:33.456789+00:00"`.

STACpopulator/populator_base.py
fmigneault: Avoid hardcoded string as much as possible to limit side-effects or having to remember (which we won't) all the places where does hardcoded string impact the code. Using predefined references makes the intention clearer.

dchandan: Well.... not quite. What you are saying is true, but only for strings that are hardcode with some information that is custom to the application. But, in this case, "DEBUG" or "INFO" are not going to mean anything else, like ever (or at least in our lifetimes :D). And if one ever has to change debugging levels in the app, searching for all instances of `logging.DEBUG` (and so on) is no different from searching for all instances of `"DEBUG"` (and so on). So, I think this is one of those things where the hardcoded string doesn't matter. However, it's changed in the code to use the more verbose form.

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
```diff
@@ -88,14 +84,21 @@ def make_parser() -> argparse.ArgumentParser:
...
+    parser.add_argument(
+        "--config",
+        type=str,
+        help=(
             "Override configuration file for the populator. "
             "By default, uses the adjacent configuration to the implementation class."
-        )
+        ),
     )
+    parser.add_argument("--debug", action="store_true", help="Set logger level to debug")
```
fmigneault, line 101: I think it should be in the parent argparser (generic one before specific implementation is called). 
`ns.debug` will not exist if using other implementations than CMIP6

STACpopulator/populator_base.py
fmigneault: Given this use case, fair enough to leave as is.

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
dchandan: Okay I've moved it. I'm not sure I like the current position much. Right now, one has to do:
```
stac-populator --debug run CMIP6_UofT stac-host catalog-url --auth-handler cookie --auth-identity ~/daccs_cookie.txt
```

I'd prefer if one could instead just do:

```
stac-populator run CMIP6_UofT stac-host catalog-url --auth-handler cookie --auth-identity ~/daccs_cookie.txt --debug
```

By the way, this might be related to #46. I noticed that there is something off with the CLI parser system you've put into place. I'll let you look into it since you put it in and hence you probably understand it better than me.

fmigneault: To allow `stac-populator run CMIP6_UofT --debug ...`, you could define the logging options as their own `log_parser = argparse.ArgumentParser(...)` and pass that `log_parser` in `parents` (https://github.com/crim-ca/stac-populator/blob/24eeb6b0b863dd5a19f1903527300903b129f9ca/STACpopulator/cli.py#L163). Need to pass it down also to `parser_maker` (https://github.com/crim-ca/stac-populator/blob/24eeb6b0b863dd5a19f1903527300903b129f9ca/STACpopulator/cli.py#L158C32-L158C44) so it can add `log_parser` to its own `parents` as well.

fmigneault (approved):
@dchandan 
Good to merge, but I let you decide if you want to try out https://github.com/crim-ca/stac-populator/pull/45#discussion_r1498326083 before or not.

dchandan:
@fmigneault 
Ok, we can do that later. Kinda pressed for time rn, so we can circle back to this later. I'll make an issue of this, so it is documented.

PR #51, opened 2024-02-23
Fixes to datacube extension helper

dchandan:
1. Some variables (e.g. `time_bnds`) do not have variable "attributes" in certain model files. As a result the extension code crashes. [8621b6d](https://github.com/crim-ca/stac-populator/commit/8621b6d1e8513de93ca5c25a62e838aacd252ec8) fixes that.
2. As per CF convention bounds variables [are not required](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#cell-boundaries) to have attributes and that they should instead be inferred from the corresponding coordinate variables. [6326908](https://github.com/crim-ca/stac-populator/commit/63169082177784cb5c31f3e050571e7b1f662300) addresses that.

An example of how the data cube extension looks like with these changes is [here](https://daccs.cs.toronto.edu/stac/collections/CMIP6_UofT/items/CMIP_CAMS_CAMS-CSM1-0_historical_r2i1p1f1_Amon_ts_gn). 

An outstanding issue is that `time_bnds` variable is still not being inferred as an auxiliary variable. This will require a hack since this variable cannot be determined to be auxiliary using the framework currently [in place](https://github.com/crim-ca/stac-populator/blob/7e7adbfc0597e8277162ae62d429f9fcb0d424dd/STACpopulator/extensions/datacube.py#L205C1-L206C1) (because the variable in the netCDF file will not contain any of the "criteria" that could be put into place for the benefit of the `is_coordinate`).

### DIFF
--- a/STACpopulator/extensions/datacube.py
+++ b/STACpopulator/extensions/datacube.py
@@ -225,6 +225,11 @@ class DataCubeHelper:
 
     def is_coordinate(self, attrs: MutableMapping[str, Any]) -> bool:
         """Return whether variable is a coordinate."""
+
+        if (desc := attrs.get("description", None)) is not None:
+            if "bounds for" in desc:
+                return True
+
         for key, criteria in self.coordinate_criteria.items():
             for criterion, expected in criteria.items():
                 if attrs.get(criterion, None) in expected:
--- a/STACpopulator/extensions/datacube.py
+++ b/STACpopulator/extensions/datacube.py
@@ -169,10 +169,8 @@ class DataCubeHelper:
                             properties = dict(
                                 type=type_.value,
                                 extent=extent,
-                                description=v.get(
-                                    "description",
-                                    v.get("long_name", criteria["standard_name"][0])
-                                ) or "",
+                                description=v.get("description", v.get("long_name", criteria["standard_name"][0]))
+                                or "",
                             )
                             if type_ == DimensionType.SPATIAL:
                                 properties["axis"] = axis
@@ -197,6 +195,8 @@ class DataCubeHelper:
                 # Some variables like "time_bnds" in some model files do not have any attributes.
                 attrs = {}
 
+            self._infer_variable_units_description(name, attrs)
+
             variables[name] = Variable(
                 properties=dict(
                     dimensions=meta["shape"],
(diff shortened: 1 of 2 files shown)


dchandan:
@huard your review would be helpful here

STACpopulator/extensions/datacube.py
```diff
@@ -191,7 +189,14 @@ def variables(self) -> dict[str, Variable]:
             if name in self.attrs["dimensions"]:
                 continue
 
-            attrs = meta["attributes"]
+            try:
+                attrs = meta["attributes"]
+            except KeyError:
+                # Some variables like "time_bnds" in some model files do not have any attributes.
+                attrs = {}
```
huard, line 196:
```suggestion
                attrs = meta.get("attributes", {})
```

STACpopulator/extensions/datacube.py
```diff
@@ -202,6 +207,25 @@ def variables(self) -> dict[str, Variable]:
...
+            related_variable = "time"
+            attrs["description"] = "bounds for the time coordinate"
+        elif name == "lat_bnds":
+            related_variable = "lat"
+            attrs["description"] = "bounds for the latitude coordinate"
+        elif name == "lon_bnds":
+            related_variable = "lon"
+            attrs["description"] = "bounds for the longitude coordinate"
+        else:
+            return
```
huard, line 222: I suspect you'd be able to harden the bound variables inference using cf-xarray. It looks for the bounds attributes of variables. See https://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries

huard: https://cf-xarray.readthedocs.io/en/latest/generated/xarray.Dataset.cf.bounds.html#xarray.Dataset.cf.bounds

huard (approved):

STACpopulator/extensions/datacube.py
huard: Now I start to remember. We didn't want to add cf-xarray as a dependency, right? So we just copied some of the logic here. Either we bring more of the cf logic into the repo, or add is as a dependency ? No strong opinion.

dchandan: Ok, then we can keep it cf-array free for now. I think this will be the path of least resistance, and time is a bit of an issue at the moment.

dchandan: @David now all the bounds variables are shown as auxiliary

fmigneault (approved):
@huard is in better position to answer regarding these attributes. Will let it up to both of you.

PR #53, opened 2024-02-26
Collection enhancements

dchandan:
- Adding ability to add collection level assets
- Adding collection assets to `CMIP6_UofT`
- Adding an end date to `CMIP6_UofT`'s temporal extent for better rendering in STAC Browser

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,8 @@
 
 <!-- insert list items of new changes here -->
 - Adding ability to add collection level assets
-- Adding collection assets to `CMIP6_UofT`
+- Adding ability to add collection level links
+- Adding collection links to `CMIP6_UofT`
 - Adding an end date to `CMIP6_UofT`'s temporal extent for better rendering in STAC Browser
 
 ## [0.6.0](https://github.com/crim-ca/stac-populator/tree/0.6.0) (2024-02-22)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,9 @@
 ## [Unreleased](https://github.com/crim-ca/stac-populator) (latest)
 
 <!-- insert list items of new changes here -->
+- Adding ability to add collection level assets
+- Adding collection assets to `CMIP6_UofT`
+- Adding an end date to `CMIP6_UofT`'s temporal extent for better rendering in STAC Browser
 
 ## [0.6.0](https://github.com/crim-ca/stac-populator/tree/0.6.0) (2024-02-22)
 
--- a/STACpopulator/implementations/CMIP6_UofT/collection_config.yml
+++ b/STACpopulator/implementations/CMIP6_UofT/collection_config.yml
@@ -5,3 +5,16 @@ keywords: ['CMIP', 'CMIP6', 'WCRP', 'Climate Change']
 license: "CC-BY-4.0"
 spatialextent: [-180, -90, 180, 90]
 temporalextent: ['1850-01-01', null]
+
+assets:
+  homepage:
+      title : Project homepage
+      href : https://wcrp-cmip.org/cmip-phase-6-cmip6/
+      description: CMIP6 project homepage
+      media_type: text/html
+  license:
+      title : License
+      href : https://raw.githubusercontent.com/WCRP-CMIP/CMIP6_CVs/main/LICENSE-CC-BY
+      description: License for CMIP6 data
+      media_type: text/plain
+
--- a/STACpopulator/implementations/CMIP6_UofT/collection_config.yml
+++ b/STACpopulator/implementations/CMIP6_UofT/collection_config.yml
@@ -4,7 +4,7 @@ description: Coupled Model Intercomparison Project phase 6
 keywords: ['CMIP', 'CMIP6', 'WCRP', 'Climate Change']
 license: "CC-BY-4.0"
 spatialextent: [-180, -90, 180, 90]
-temporalextent: ['1850-01-01', null]
+temporalextent: ['1850-01-01', '2500-01-01']
 
 assets:
   homepage:
--- a/STACpopulator/implementations/CMIP6_UofT/collection_config.yml
+++ b/STACpopulator/implementations/CMIP6_UofT/collection_config.yml
@@ -6,15 +6,15 @@ license: "CC-BY-4.0"
 spatialextent: [-180, -90, 180, 90]
 temporalextent: ['1850-01-01', '2500-01-01']
 
-assets:
+links:
   homepage:
+      rel: about
       title : Project homepage
-      href : https://wcrp-cmip.org/cmip-phase-6-cmip6/
-      description: CMIP6 project homepage
+      target : https://wcrp-cmip.org/cmip-phase-6-cmip6/
       media_type: text/html
   license:
+      rel: license  
       title : License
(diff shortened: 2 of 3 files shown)


STACpopulator/implementations/CMIP6_UofT/collection_config.yml
```diff
@@ -4,4 +4,17 @@ description: Coupled Model Intercomparison Project phase 6
...
+  homepage:
+      title : Project homepage
+      href : https://wcrp-cmip.org/cmip-phase-6-cmip6/
+      description: CMIP6 project homepage
+      media_type: text/html
+  license:
+      title : License
+      href : https://raw.githubusercontent.com/WCRP-CMIP/CMIP6_CVs/main/LICENSE-CC-BY
+      description: License for CMIP6 data
+      media_type: text/plain
```
fmigneault, line 19: This should be added in `links` with `rel: license` instead of `assets`: 
https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#license

STACpopulator/implementations/CMIP6_UofT/collection_config.yml
```diff
@@ -4,4 +4,17 @@ description: Coupled Model Intercomparison Project phase 6
...
-temporalextent: ['1850-01-01', null]
\ No newline at end of file
+temporalextent: ['1850-01-01', '2500-01-01']
+
+assets:
+  homepage:
+      title : Project homepage
+      href : https://wcrp-cmip.org/cmip-phase-6-cmip6/
+      description: CMIP6 project homepage
+      media_type: text/html
```
fmigneault, line 14: Seems more appropriate to use `links` with `rel: about`. 
https://www.iana.org/assignments/link-relations/link-relations.xhtml

fmigneault (commented):
STAC Assets are usually intended for referencing to data or metadata directly relating to the data (eg: some configuration file to interpret it, alternate representations, spec-sheet of the satellite or the convention names of the attributes). 

For contextual metadata around the item, such as the CMIP6 project and license, the `links` section seems more appropriate, using standard IANA relation types known even by non-STAC JSON parsers. 

The feature (STAC Assets in Collections) is valid by itself, but the CMIP6 collection config should instead use `links` with the specific changes provided.

CHANGES.md
```diff
@@ -3,6 +3,9 @@
 ## [Unreleased](https://github.com/crim-ca/stac-populator) (latest)
 
 <!-- insert list items of new changes here -->
+- Adding ability to add collection level assets
+- Adding collection assets to `CMIP6_UofT`
+- Adding an end date to `CMIP6_UofT`'s temporal extent for better rendering in STAC Browser
```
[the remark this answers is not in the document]
dchandan: The "until present" is not ideal because it's not true. The final date for CMIP6 data is known, the maximum a simulation in the archive goes up to is year 2500. When I originally had "null" I was expecting something like STAC browser to show an undetermined/open end date, not something wrong like "until present".

STACpopulator/implementations/CMIP6_UofT/collection_config.yml
dchandan: So, I took this from your example: https://hirondelle.crim.ca/stac/collections/EuroSAT-subset-train. You have license in assets. Is that wrong?

STACpopulator/implementations/CMIP6_UofT/collection_config.yml
dchandan: Again it seemed from my examination of your implementation that you were using about like links in assets.

huard (approved):
Temporal extent is good for me. 
Don't feel qualified to weigh in on the asset vs link discussion.

STACpopulator/implementations/CMIP6_UofT/collection_config.yml
fmigneault: Yes. Looks like a misuse.

The `license` in `links` is specifically recommended by STAC. 
It is also commonly used, for example: https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-c2-l2

CHANGES.md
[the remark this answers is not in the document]
fmigneault: If this is a specific date imposed by CMIP6, then it is fine. I'm not familiar with that granular level of details about it.

STACpopulator/implementations/CMIP6_UofT/collection_config.yml
dchandan: @fmigneault was this a misuse too?

STACpopulator/implementations/CMIP6_UofT/collection_config.yml
dchandan: I've now updated the assets to links.

STACpopulator/implementations/CMIP6_UofT/collection_config.yml
fmigneault: Yes. This should also be in `links`.

STACpopulator/populator_base.py
```diff
@@ -135,13 +135,50 @@ def create_stac_collection(self) -> dict[str, Any]:
...
+
+        # Add any assets if provided in the config
+        if "assets" in self._collection_info:
+            self._collection_info["assets"] = self.__make_collection_assets()
+
+        # Construct links if provided in the config
+        if "links" in self._collection_info:
+            collection_links = self.__make_collection_links()
+            # need to remove the links item other constructing a collection object fails
+            _ = self._collection_info.pop("links")
```
fmigneault, line 147: typo "other" -> "otherwise" ?

I think it would make more sense to delegate all the checks to `self.__make_collection_links` to avoid spliting the logic in two places.
i.e.: 

```python
collection_links = self.__make_collection_links()  # called directly

def __make_collection_links(self):
    links = self._collection_info.pop("links", [])
    for link_name, link_info in self._collection_info["links"].items():
        links.append(pystac.Link(**link_info))
    return links
```

Same idea for `"assets"`.

fmigneault (commented):
Recommended change, but not blocking.

STACpopulator/populator_base.py
dchandan: Yeah, that's a good idea. Let me rearrange it that way.

STACpopulator/implementations/CMIP6_UofT/collection_config.yml
```diff
@@ -4,4 +4,17 @@ description: Coupled Model Intercomparison Project phase 6
...
+  homepage:
+      rel: about
+      title : Project homepage
+      target : https://wcrp-cmip.org/cmip-phase-6-cmip6/
+      media_type: text/html
+  license:
+      rel: license  
+      title : License
+      target : https://raw.githubusercontent.com/WCRP-CMIP/CMIP6_CVs/main/LICENSE-CC-BY
+      media_type: text/plain
```
fmigneault, line 19: Why use a dict instead of a list if the link name is dropped anyway?

dchandan: That's a good idea. Done.

dchandan:
Merging, since it looks like we are good based on your earlier comment.

PR #54, opened 2024-02-27
use cf-xarray heuristic to identify bound variables. Add test

huard:
@dchandan Replace your inference logic by the same logic used in cf-xarray

Added test

### DIFF
--- a/STACpopulator/extensions/datacube.py
+++ b/STACpopulator/extensions/datacube.py
@@ -184,6 +184,7 @@ class DataCubeHelper:
     def variables(self) -> dict[str, Variable]:
         """Return Variable objects required for Datacube extension."""
         variables = {}
+        bounds = self.bounds()
 
         for name, meta in self.attrs["variables"].items():
             if name in self.attrs["dimensions"]:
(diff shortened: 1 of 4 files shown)


fmigneault (commented):
Code looks fine. 
Just extra test that oculd be relevant.

tests/test_cmip6_datacube.py
```diff
@@ -35,3 +35,22 @@ def test_datacube_helper():
...
+    attrs["access_urls"] = {"HTTPServer": "http://example.com"}
+    item = CMIP6Helper(attrs, GeoJSONPolygon).stac_item()
+
+    dc = DataCubeHelper(attrs)
+    dc_ext = DatacubeExtension.ext(item, add_if_missing=True)
+    dc_ext.apply(dimensions=dc.dimensions, variables=dc.variables)
+
+    p = dc_ext.properties
+    assert set(['time', 'lat', 'lon']) == set(p['cube:dimensions'].keys())
+    assert p["cube:variables"]["lon_bnds"]["unit"] == "degrees_east"
```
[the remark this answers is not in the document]
huard: Done

STACpopulator/extensions/datacube.py
```diff
@@ -192,44 +193,47 @@ def variables(self) -> dict[str, Variable]:
...
+                # Bounds are auxiliary variables
+                dtype = VariableType.AUXILIARY.value
+
+                # We can safely assume that the bounds variable has the same units as the variable it bounds.
+                if "units" not in attrs:
+                    if (u := self.attrs["variables"][bounds[name]].get("attributes", {}).get("units")) is not None:
+                        attrs["units"] = u
+
+            elif self.is_coordinate(attrs):
+                # Using the CF-xarray heuristics to determine if variable is a coordinate.
```
dchandan, line 206: Is this if condition ever entered, given the presence of the `continue` statement on line 191?

STACpopulator/extensions/datacube.py
```diff
@@ -192,44 +193,47 @@ def variables(self) -> dict[str, Variable]:
...
+                dtype = VariableType.AUXILIARY.value
+            else:
+                dtype = VariableType.DATA.value
 
             variables[name] = Variable(
                 properties=dict(
                     dimensions=meta["shape"],
-                    type=VariableType.AUXILIARY.value if self.is_coordinate(attrs) else VariableType.DATA.value,
+                    type=dtype,
                     description=attrs.get("description", attrs.get("long_name", "")),
```
dchandan, line 215: I think we should add a description to the bounds variable too. Something like "bounds for the time coordinate" as in #51. We can construct an appropriate description string in the `if-elif-else` block before this and use that here.

dchandan (approved):
Looks good. Just a minor suggestion, but not blocking.

STACpopulator/extensions/datacube.py
dchandan: Okay, I guess the two are complementary and not identical.

huard: Correct.

STACpopulator/extensions/datacube.py
huard: So I wanted to hear your thoughts on this. 

1. What's the rationale for "augmenting" the metadata attributes when creating the STAC items?
1. Is the datacube extension logic the appropriate place to do this? 
1. Will it scale well to other missing infos?
1. Are we worried about a "surprise" factor coming from having information in the STAC item that does not reflect what people will find in the actual dataset?

dchandan: Good questions, here's what I think:

1. I think the rationale would be that the STAC item should be as informative as reasonably possible. If, at the STAC item generation phase adding information that doesn't exist in the file metadata achieves that goal, then I don't see why not. In general, there is no principle/constraint requiring one-to-one correspondence between metadata in a STAC representation and that in the underlying asset. Here, we just use metadata from the CMIP6 files to guide the generation of STAC items, but even if the CMIP6 files were poorly constructed with relevant metadata, we could still generate the STAC items by programmatically adding the relevant metadata at item generation time.
2. I really don't know what is the purpose of the datacube extension. I have never used it, and I don't envision using it. But from looking at the data it represents, it seems it just described the dimensions and variable available in an asset, and if that is the case, then we definitely can use the extension to add a description to a variable that is listed by the extension.
3. I don't see why not. Going back to point 1, I think it's all about including whatever one feels is necessary for describing the data variables and dimensions.
4. No, I am not. Going back to point 1, there is no constraint about a one-to-one correspondence between metadata described in the STAC representation and that in an underlying file. Consider the case of some obscure dataset where the underlying .nc files do not have any metadata whatsoever (on variable level or global). There is nothing preventing me from appropriately describing the metadata in the STAC layer and make the dataset available via the catalog. In this case, none of the metadata in the STAC representation would be available in the file itself, and that is perfectly OK.

huard: Ok, I'm happy adopting your perspective on 1 and 4. I'm less convinced about 2 and 3 and I believe we should spend some time thinking about the "metadata augmentation architecture" before writing too much code. I'm happy to add the description you mentioned, but I don't think this should serve as a template to add other kinds of information. 

My discomfort comes from mixing process logic (NcML -> STAC) with thematic knowledge (metadata fixes) in the same code base. I think we can get the process logic fairly robust and maintenance should be minimal. The thematic knowledge will always grow with new datasets. Mixing both has the potential to generate maintenance work in the future.

fmigneault (approved):

PR #58, opened 2024-10-07
Do not require pyessv archive

mishaschwartz:
The pyessv-archive is required to run any stac-populator CLI commands but should only be required to run the CMIP6 populator. 

This PR allows the CLI to work as intended even if pyessv-archive is not installed. Instead of exiting with an error, it will show a warning and continue to provide all other functionality that does not require pyessv-archive to run.

Current behaviour:

```
$ stac-populator --help
...
...
STACpopulator.exceptions.ExtensionLoadError: /home/.../.esdoc/pyessv-archive directory does not exists
```

After this PR behaviour:

```
$ stac-populator --help
.../stac-populator/STACpopulator/cli.py:159: UserWarning: Could not load extension CMIP6_UofT because of error /home/mschwa/.esdoc/pyessv-archive directory does not exists
  warnings.warn(f"Could not load extension {populator_name} because of error {e}")
usage: stac-populator [-h] [--version] [--debug] {run} ...
```

This PR also updates the README.md to clarify the dependency on pyessv-archive and fixes some minor typos.

Finally, it changes the `PYESSV_ARCHIVE_DIR` variable in the Makefile to `PYESSV_ARCHIVE_HOME` which is the variable used by the [pyessv](https://github.com/ES-DOC/pyessv) project. This means that users who wish to install the archive in a non-default location only need to set one variable, not two.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "pyessv-archive"]
-	path = pyessv-archive
-	url = https://github.com/dchandan/pyessv-archive
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 * Adding collection links to `CMIP6_UofT`
 * Adding an end date to `CMIP6_UofT`'s temporal extent for better rendering in STAC Browser
 * Updates to datacube extension helper routines for `CMIP6_UofT`.
+* Make pyessv-archive a requirement for *only* the cmip6 implementation instead of for the whole CLI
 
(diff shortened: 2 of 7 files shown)


huard (approved):

huard:
Looks good. FYI, trying to convince maintainers of CORDEX and CMIP CVs to host a jsonschema built on their CVs. Worst case, we can host those ourselves, and replace the validation using pyessv with a simple validate from jsonschema.

fmigneault (approved):

PR #60, opened 2024-10-09
fix CLI and logging

mishaschwartz:
While experimenting with the CLI I realized that the implementations could not be invoked through the CLI due to an import error in the `logging.py` script.

This PR fixes that error and adds tests for the CLI code itself to ensure that these errors are detected earlier. Testing required refactoring the CLI code in order to test it properly. While refactoring the code, I also cleaned it up a bit and enforced some restrictions on return types from functions.

Finally, there was no mechanism to specify an alternative place to write log files, so I added one. 

Summary of changes:

* Fix bug where logger setup failed
* Simplify CLI argument constructor code (for cleaner and more testable code)
* Add tests for CLI and implementations when invoked through the CLI
* Refactored code dealing with requests and authentication to the `requests.py` file
* Add `--log_file` command line option to specify a non-default location to write log files to

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,8 +14,8 @@
 * Fix bug where logger setup failed
 * Simplify CLI argument constructor code (for cleaner and more testable code)
 * Add tests for CLI and implementations when invoked through the CLI
-* Refactored code dealing with requests and authentication to the `requests.py` file
-* Add `--log_file` command line option to specify a non-default location to write log files to
+* Refactored code dealing with requests and authentication to the `STACpopulator/requests.py` file
+* Add `--log-file` command line option to specify a non-default location to write log files to
 
 ## [0.6.0](https://github.com/crim-ca/stac-populator/tree/0.6.0) (2024-02-22)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,11 @@
 * Adding an end date to `CMIP6_UofT`'s temporal extent for better rendering in STAC Browser
 * Updates to datacube extension helper routines for `CMIP6_UofT`.
 * Make pyessv-archive a requirement for *only* the cmip6 implementation instead of for the whole CLI
+* Fix bug where logger setup failed
+* Simplify CLI argument constructor code (for cleaner and more testable code)
+* Add tests for CLI and implementations when invoked through the CLI
+* Refactored code dealing with requests and authentication to the `requests.py` file
+* Add `--log_file` command line option to specify a non-default location to write log files to
 
 ## [0.6.0](https://github.com/crim-ca/stac-populator/tree/0.6.0) (2024-02-22)
 
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -23,7 +23,7 @@ def add_parser_args(parser: argparse.ArgumentParser) -> dict[str, Callable]:
     )
     parser.add_argument("--debug", action="store_const", const=logging.DEBUG, help="set logger level to debug")
     parser.add_argument(
-        "--log_file", help="file to write log output to. By default logs will be written to the current directory."
+        "--log-file", help="file to write log output to. By default logs will be written to the current directory."
     )
     commands_subparser = parser.add_subparsers(
         title="command", dest="command", description="STAC populator command to execute.", required=True
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -1,108 +1,19 @@
 import argparse
-import glob
+import functools
 import importlib
 import logging
-import os
 import sys
+from types import ModuleType
 import warnings
-from datetime import datetime
-from http import cookiejar
-from typing import Callable, Optional
+from datetime import datetime, timezone
+from typing import Callable
 
-import requests
-from requests.auth import AuthBase, HTTPBasicAuth, HTTPDigestAuth, HTTPProxyAuth
-from requests.sessions import Session
-
-from STACpopulator import __version__
+from STACpopulator import __version__, implementations
 from STACpopulator.exceptions import STACPopulatorError
 from STACpopulator.logging import setup_logging
 
-POPULATORS = {}
 
-
-class HTTPBearerTokenAuth(AuthBase):
-    def __init__(self, token: str) -> None:
-        self._token = token
-
-    def __call__(self, r: requests.PreparedRequest) -> requests.PreparedRequest:
-        r.headers["Authorization"] = f"Bearer {self._token}"
-        return r
-
-
-class HTTPCookieAuth(cookiejar.MozillaCookieJar):
-    """
-    Employ a cookie-jar file for authorization.
-
-    Examples of useful command:
-
-    .. code-block:: shell
-
-        curl --cookie-jar /path/to/cookie-jar.txt [authorization-provider-arguments]
-
-        curl \
-            -k \
-            -X POST \
-            --cookie-jar /tmp/magpie-cookie.txt \
-            -d '{"user_name":"...","password":"..."}' \
-            -H 'Accept:application/json' \
-            -H 'Content-Type:application/json' \
-            'https://{hostname}/magpie/signin'
-
-    .. note::
-        Due to implementation details with :mod:`requests`, this must be passed directly to the ``cookies``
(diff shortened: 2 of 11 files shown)


huard (approved):
Gave a quick look, looks good from my end and takes one step toward generalization.

STACpopulator/implementations/NEX_GDDP_UofT/add_NEX-GDDP.py
fmigneault, line 1: What's the deal with this one? No intention from UofT to offer it anymore?

STACpopulator/cli.py
```diff
@@ -1,217 +1,67 @@
...
-        prog=f"{parser.prog} {run_cmd_parser.prog}",
-        parents=[run_cmd_parser],
-        formatter_class=run_cmd_parser.formatter_class,
-        usage=run_cmd_parser.usage,
-        add_help=False,
-        help=run_cmd_parser.description,
-        description=run_cmd_parser.description,
+    parser.add_argument("--debug", action="store_const", const=logging.DEBUG, help="set logger level to debug")
+    parser.add_argument(
+        "--log_file", help="file to write log output to. By default logs will be written to the current directory."
```
fmigneault, line 26: Small nitpick.  Use `--log-file`. CLIs usually employ hyphens.

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
```diff
@@ -123,13 +123,14 @@ def runner(ns: argparse.Namespace) -> Optional[int] | NoReturn:
             ns.stac_host, data_loader, update=ns.update, session=session, config_file=ns.config, log_debug=ns.debug
         )
         c.ingest()
+    return 0
 
 
-def main(*args: str) -> Optional[int]:
-    parser = make_parser()
+def main(*args: str) -> int:
+    parser = argparse.ArgumentParser()
```
fmigneault, line 130: If one calls `python STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py`, this will now fail. 
The previous `make_parser` approach allowed to do either approach. Is it only missing a `add_parser_args` call to accomplish the same?

STACpopulator/implementations/DirectoryLoader/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .crawl_directory import add_parser_args, runner
```
fmigneault, line 1: use absolute import

STACpopulator/logging.py
```diff
@@ -1,6 +1,6 @@
 import datetime as dt
 import json
-import logging
+import logging.config
```
fmigneault, line 3: There is (at least) `logging.getLevelName` reference in there, so `import logging` is needed as well.

STACpopulator/requests.py
```diff
@@ -0,0 +1,60 @@
...
+        if namespace.auth_handler == "basic":
+            session.auth = HTTPBasicAuth(usr, pwd)
+        elif namespace.auth_handler == "digest":
+            session.auth = HTTPDigestAuth(usr, pwd)
+        else:
+            session.auth = HTTPProxyAuth(usr, pwd)
+    elif namespace.auth_handler == "bearer":
+        session.auth = HTTPBearerTokenAuth(namespace.auth_identity)
+    elif namespace.auth_handler == "cookie":
+        session.cookies = cookiejar.MozillaCookieJar(namespace.auth_identity)
```
fmigneault, line 59: Curious if `MozillaCookieJar` is just a poorly named class, or whether that actually injects anything Mozilla related that could cause side effects in requests?  For example, Magpie/Twitcher return different responses contents by default based on the detected `User-Agent`.

CHANGES.md
```diff
@@ -11,6 +11,11 @@
 * Adding an end date to `CMIP6_UofT`'s temporal extent for better rendering in STAC Browser
 * Updates to datacube extension helper routines for `CMIP6_UofT`.
 * Make pyessv-archive a requirement for *only* the cmip6 implementation instead of for the whole CLI
+* Fix bug where logger setup failed
+* Simplify CLI argument constructor code (for cleaner and more testable code)
+* Add tests for CLI and implementations when invoked through the CLI
+* Refactored code dealing with requests and authentication to the `requests.py` file
+* Add `--log_file` command line option to specify a non-default location to write log files to
```
fmigneault, line 18: update as per other comment

CHANGES.md
```diff
@@ -11,6 +11,11 @@
 * Adding an end date to `CMIP6_UofT`'s temporal extent for better rendering in STAC Browser
 * Updates to datacube extension helper routines for `CMIP6_UofT`.
 * Make pyessv-archive a requirement for *only* the cmip6 implementation instead of for the whole CLI
+* Fix bug where logger setup failed
+* Simplify CLI argument constructor code (for cleaner and more testable code)
+* Add tests for CLI and implementations when invoked through the CLI
+* Refactored code dealing with requests and authentication to the `requests.py` file
```
fmigneault, line 17: Add the `STACpopulator` prefix. Too easy to confuse `requests.py` with the "Python `requests` package".

STACpopulator/implementations/NEX_GDDP_UofT/add_NEX-GDDP.py
mishaschwartz: The file is empty. We offer that dataset but there's nothing here. 

If there is content for this file in another branch that never made it to a PR, I don't see it. :shrug:

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
mishaschwartz: Yup it's missing. I'll add it in. 

Personally I don't love that there are multiple interfaces for running this script and I really think that we should discourage running these files directly. But that's a discussion for another time.

mishaschwartz: (this is the reason that I didn't include more tests for running these scripts directly from the command line)

STACpopulator/logging.py
mishaschwartz: `import logging.config` adds the `logging` module to the local namespace as well (if it didn't we would see a failing test). Adding an additional `import logging` is redundant.

STACpopulator/requests.py
mishaschwartz: It just uses the mozilla file format (see: https://docs.python.org/3/library/http.cookiejar.html#http.cookiejar.MozillaCookieJar). There are a few warnings about usage in the docs there but if we're concerned we should address them in another PR (this PR maintains the previous behaviour).

STACpopulator/implementations/NEX_GDDP_UofT/add_NEX-GDDP.py
fmigneault: That's fine. Just wanted to make sure.

STACpopulator/cli.py
```diff
@@ -1,217 +1,67 @@
...
+    commands_subparser = parser.add_subparsers(
+        title="command", dest="command", description="STAC populator command to execute.", required=True
+    )
+    run_parser = commands_subparser.add_parser("run", description="Run a STACpopulator implementation")
+    populators_subparser = run_parser.add_subparsers(
+        title="populator", dest="populator", description="Implementation to run."
+    )
+    for implementation_module_name, module in implementation_modules().items():
+        implementation_parser = populators_subparser.add_parser(implementation_module_name)
+        module.add_parser_args(implementation_parser)
```
huard, line 37: Starting to look into this, and I'm not really fond of the logic depending on a function named "add_parser_args" being present in the module. I don't have a better alternative yet though.

fmigneault: The previous approach assumed a `make_parser` instead. The logic just changed the name to represent what the function does, but has similar dependencies.

The only workaround I can think of is to define an abstract `ArgParser` class with a `add_parser_args` method, and each implementation must define their `ArgParser` inheriting from it. The CLI could then look up the modules and filter by `issubclass` of `ArgParser` to find relevant references. That being said, it just defines an explicit structure, but the result would be the same.

PR #64, opened 2024-10-11
Cordex extension and trying to build on higher level abstractions

huard:
I added a CMIP6-CORDEX extension and implementation, trying to create base classes that would simplify the addition of other extensions. 

This simplifies a bit the implementation part, but you'll see that there is still some boilerplate code we could do without on the implementation side. 

The main change is that I created a generic `THREDDSCatalogDataModel`. Extensions then only have to define the data model for their properties, and how to construct a unique ID. If a jsonschema is provided, then it will be used to validate the incoming data. I've disabled the validation done at the STAC extension level for now (see below).

I've struggled a bit with the role of the jsonschema here. In climate science, this is not a very popular tool. Even if scientific schemas appeared, we'd have to embed them into a STAC specific schema. You'll see that I've created a schema directory with the CORDEX schema for global attributes, but this is not a STAC schema per say. A STAC schema would embed those attributes into a `property` object, accompanied by a `type`. I didn't know how to embed a schema into another, that's why I disabled the extension schema validation. 

To try it: 
```bash
stac-populator run Ouranos_CMIP6-CORDEX http://localhost:8880/stac https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/disk2/ouranos/CORDEX/CMIP6/DD/NAM-12/OURANOS/MPI-ESM1-2-LR/ssp370/r1i1p1f1/CRCM5/v1-r1/day/tas/v20231208/catalog.html
```

If you think this kind of abstraction is useful, I could port those changes to the CMIP6 case in another PR.


STACpopulator/implementations/Ouranos_CMIP6-CORDEX/add_CORDEX6.py
```diff
@@ -0,0 +1,75 @@
...
+LOGGER = logging.getLogger(__name__)
+import argparse
+from typing import Any
+from STACpopulator.populator_base import STACpopulatorBase
+from STACpopulator.extensions.cordex6 import Cordex6DataModel
+
+
+class CORDEX_STAC_Populator(STACpopulatorBase):
+    data_model = Cordex6DataModel
+    item_geometry_model = None  # Unnecessary, but kept for consistency
```
fmigneault, line 29: This is defined for the `CMIP6populator`: 

```python
class CMIP6populator(STACpopulatorBase):
    item_properties_model = CMIP6Properties
    item_geometry_model = GeoJSONPolygon
```

And `data_model = Cordex6DataModel` basically offers: 

```python
Cordex6DataModel.properties == CordexCmip6  #  -> just like CMIP6Properties
Cordex6DataModel -> THREDDSCatalogDataModel.geometry  # -> just like item_geometry_model
```

I'm wondering if there's any duplication of the intended use of these properties?

STACpopulator/extensions/cordex6.py
```diff
@@ -0,0 +1,95 @@
...
+            "variable_id",
+            "domain_id",
+        ]
+        name = "_".join(getattr(self.properties, k) for k in keys)
+        return name
+
+
+
+# TODO: Remove before merging
+def get_test_data():
```
fmigneault, line 74: move to `tests` dir

STACpopulator/extensions/cordex6.py
```diff
@@ -0,0 +1,95 @@
...
+from datetime import datetime
+
+from importlib import reload
+import STACpopulator.extensions.base
+reload(STACpopulator.extensions.base)
+from STACpopulator.extensions.base import THREDDSCatalogDataModel, DataModel
+
+
+# This is generated using datamodel-codegen + manual edits
+class CordexCmip6(DataModel):
```
fmigneault, line 14: Since the model is generated from the schema, why is the `@model_validator` needed to load and validate the JSON schema?

I'm not seeing the subtlety from static code analysis.

fmigneault (commented):
About the "embedded schema", I think an extension could be defined as such (using YAML for short, but convert to JSON for applying it): 

```yaml
type: object
required:
  - type
  - properties
properties:
  type: 
    const: Feature
  properties:
    $ref: "STACpopulator/extensions/schemas/cordex6/cmip6-cordex-global-attrs-schema.json"
```

Then, you can reuse the JSON schema on its own or as STAC extension definition. 

As for the PR itself, I have a strong sensation that `THREDDSCatalogDataModel` is essentially trying to accomplish what the "helpers" were trying to do (but missing some interface to connect the dots). 

It is a bit hard to analyze the code path with all the abstractions involved. So, if I misinterpreted something in my comments, please let me know. 


## other todos

- [x] Need to add `Ouranos_CMIP6-CORDEX` to the table in the README. 
- [ ] Update changelog

.gitignore
```diff
@@ -22,7 +22,6 @@ build
 
 ## Logs
 *.jsonl
-*.json
```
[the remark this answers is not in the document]
huard: Didn't think of that, thanks.

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,330 @@
...
+    Subclass this with attributes.
+    """
+    # Ideally, the catalog properties would be described by a jsonschema.
+    _prefix: str = PrivateAttr()
+
+    # URI of the json schema to validate against.
+    _schema_uri: FilePath = PrivateAttr(None)
+
+    # List of properties not meant to be validated by json schema.
+    _schema_exclude: list[str] = PrivateAttr([])
```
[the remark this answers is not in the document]
huard: I don't think so, because those were fields I wanted to exclude from the schema validation, but not from the model dump. I was thinking of a case where the schema is strictly prohibiting extra attributes, but I realize this might be a very edgy corner case.

STACpopulator/populator_base.py
```diff
@@ -221,3 +226,55 @@ def ingest(self) -> None:
 
             counter += 1
             LOGGER.info(f"Processed {counter} data items. {failures} failures")
+            
+
+
+class THREDDSRunner:
```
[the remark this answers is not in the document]
huard: Pending. I started working on abstracting the "runner" logic, but realized it would break the current CMIP implementation, so decided to focus only on the extension part. 

I can remove it from the PR and come back to it later.

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,330 @@
...
+    model_config = ConfigDict(populate_by_name=True, extra="ignore")
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_jsonschema(cls, data):
+        """Validate the data model against the json schema, if given."""
+        # Load schema
+        uri =  cls._schema_uri.default
+        if uri is not None:
+            schema = json.load(open(uri))
```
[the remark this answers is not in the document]
huard: I was unsure how to deal with references within a schema if it was not local.

STACpopulator/implementations/Ouranos_CMIP6-CORDEX/add_CORDEX6.py
huard: Yes, because I didn't want to break the CMIP extension and implementation just yet. My idea was to try to generalize the CORDEX example, get a sense of where this is going, and once we're happy, then bring the changes to CMIP6.

STACpopulator/extensions/cordex6.py
huard: It's slimmed down version of the schema without the actual CV validation. The schema includes enums with the CVs, while the pydantic.DataModel does not. 

This is a question I struggled with. I felt it didn't make a lot of sense to duplicate the jsonschema validation in pydantic. On the other hand, relying only on the schema and not even seeing the attributes in the code felt obscure and not admin friendly. So I thought it would be useful to have a `pydantic DataModel` layer where you can add attributes to the data model, and exclude some that are in the schema but you don't want in the STAC item.

STACpopulator/implementations/Ouranos_CMIP6-CORDEX/add_CORDEX6.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def create_stac_item(self, item_name: str, item_data: dict[str, Any]) -> dict[str, Any]:
+        dm = self.data_model.from_data(item_data)
+        return dm.stac_item()
```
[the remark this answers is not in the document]
huard: Note that that `THREDDSCatalogDataModel` automatically applies the `datacube` and `thredds` extension.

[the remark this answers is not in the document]
huard: One issue I see with this is that the extension helpers have different `__init__` requirements. So either the helpers know how to parse the input data, or the object instantiating them provides that logic.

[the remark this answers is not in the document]
fmigneault: We could have an in-between solution. 
The `item_helpers` list could define instances rather than type references: 
```python
item_helpers = [
  HelperWithoutArg(),
  THREDDSHelper(["<url>"]),
]
```
Anything that can be supplied at init would be created right away, and the STAC `item` objects would be obtained during the `apply(item)` call. 

I don't think there are any cases where the helpers would be missing references limiting this approach, but to investigate...

STACpopulator/extensions/cordex6.py
fmigneault: I wonder if that is an issue with datamodel-codegen, or an option to provide? 
Normally, the enums should be possible using `Literal` type with pydantic. 

I think it makes sense to have the `DataModel` auto-generated from schema to provide the attributes. It's easier to manipulate by users used to Python but not so much JSON schema.

huard: Yes, it's definitely possible. I just didn't include the Literals in the python code.

STACpopulator/implementations/Ouranos_CMIP6-CORDEX/add_CORDEX6.py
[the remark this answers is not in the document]
huard: Not sure I follow. You need the data to create instances of the helpers.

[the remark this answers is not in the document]
huard: What we could do is something like this:
```python
    @classmethod
    def from_data(cls, data):
        """Instantiate class from data provided by THREDDS Loader.
        """
        # This is where we match the Loader's output to the STAC item and extensions inputs. If we had multiple
        # loaders, that's probably the only thing that would be different between them.
        return cls(data=data,
                   start_datetime=data["groups"]["CFMetadata"]["attributes"]["time_coverage_start"],
                   end_datetime=data["groups"]["CFMetadata"]["attributes"]["time_coverage_end"],
                   geometry=ncattrs_to_geometry(data),
                   bbox=ncattrs_to_bbox(data),
                   properties=data["attributes"],
                   )

    @model_validator(mode="before")
    @classmethod
    def datacube_helper(cls, data):
        """Validate the DataCubeHelper."""
        data["datacube"] = DataCubeHelper(data['data'])
        return data

    @model_validator(mode="before")
    @classmethod
    def thredds_helper(cls, data):
        """Validate the DataCubeHelper."""
        data["thredds"] = THREDDSHelper(data['data']["access_urls"])
        return data
```

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,360 @@
...
+    extensions : list[str]
+        Name of the class attributes that point to STAC extension helper classes. Those extension classes should have an `apply` method.
+    """
+    # STAC item properties
+    geometry: AnyGeometry | None
+    bbox: list[float]
+    start_datetime: datetime
+    end_datetime: datetime
+
+    extensions: list = []
```
mishaschwartz, line 129: Is there a reason these can't be a list of the extensions themselves? Why have this as a list of strings referring to class methods?

STACpopulator/extensions/datacube.py
```diff
@@ -248,3 +248,10 @@ def temporal_extent(self) -> MutableSequence[str]:
         start_datetime = cfmeta["time_coverage_start"]
         end_datetime = cfmeta["time_coverage_end"]
         return [start_datetime, end_datetime]
+
```
mishaschwartz, line 251: Should this class be a subclass of `ExtensionHelper`?

STACpopulator/implementations/Ouranos_CMIP6-CORDEX/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .add_CORDEX6 import add_parser_args, runner
```
mishaschwartz, line 1: Let's make this an absolute import instead of relative

STACpopulator/extensions/thredds.py
```diff
@@ -132,3 +132,18 @@ def links(self) -> list[pystac.Link]:
         url = self.access_urls[ServiceType.httpserver]
         link = magpie_resource_link(url)
         return [link]
+
```
mishaschwartz, line 135: Same here... should this be a subclass of `ExtensionHelper`

STACpopulator/implementations/Ouranos_CMIP6-CORDEX/add_CORDEX6.py
```diff
@@ -0,0 +1,62 @@
+import logging
+from requests.sessions import Session
+
+from STACpopulator.requests import add_request_options, apply_request_options
+from STACpopulator.input import ErrorLoader, THREDDSLoader
+
+LOGGER = logging.getLogger(__name__)
```
mishaschwartz, line 7: global variable assignments should be after the import statements

STACpopulator/implementations/Ouranos_CMIP6-CORDEX/add_CORDEX6.py
```diff
@@ -0,0 +1,62 @@
...
+    parser.add_argument(
+        "--config",
+        type=str,
+        help=(
+            "Override configuration file for the populator. "
+            "By default, uses the adjacent configuration to the implementation class."
+        ),
+    )
+    add_request_options(parser)
+
```
mishaschwartz, line 45: With the recent update you should now include:

```python
add_logging_options(parser)
```

And you'll need to add at the top:

```python
from STACpopulator.log import add_logging_options
```

tests/test_cordex.py
```diff
@@ -0,0 +1,24 @@
+from STACpopulator.extensions.cordex6 import Cordex6DataModel
+
+
+def get_test_data():
+    import requests
+    from siphon.catalog import TDSCatalog
+    import xncml
+    from STACpopulator.stac_utils import numpy_to_python_datatypes
+
+    cat = TDSCatalog("https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/disk2/ouranos/CORDEX/CMIP6/DD/NAM-12/OURANOS/MPI-ESM1-2-LR/ssp370/r1i1p1f1/CRCM5/v1-r1/day/tas/v20231208/catalog.html")
```
mishaschwartz, line 10: Can we not use a real dataset that has to be remotely accessed in these tests? It would be preferable to mock the data so that these tests can be run offline and they won't be dependent on the PAVICS THREDDS server to pass.

STACpopulator/extensions/datacube.py
huard: Yes, I think it would be good to have a clear hierarchy, but I didn't want to break too much stuff in one PR.

STACpopulator/extensions/thredds.py
huard: ExtensionHelper is a pydantic class, while the Datacube and THREDDS helpers are not. I agree it should be harmonized, but I suggest we do this in another PR.

tests/test_cordex.py
huard: Agree, I'll make a static version.

STACpopulator/extensions/base.py
fmigneault: I think it is better to let the helpers add them automatically, as they might change over time (e.g.: new extension version with modified attributes).

mishaschwartz: Sorry, I'll try to explain a bit better (or maybe I'm not understanding the issue). 

Right now we do something like this:

```python
class BaseSTAC:
    ...
    def stac_item(self) -> "pystac.Item":
        ...
        for ext in self.extensions:
            getattr(self, ext).apply(item)

class THREDDSCatalogDataModel(BaseSTAC):
    ...
    properties: ExtensionHelper
    datacube: DataCubeHelper
    thredds: THREDDSHelper
    ...
    extensions: list = ["properties", "datacube", "thredds"]
```

Why can't we do this?

```python
class BaseSTAC:
    ...
    def stac_item(self) -> "pystac.Item":
        ...
        for ext in self.extensions:
            ext.apply(item)

class THREDDSCatalogDataModel(BaseSTAC):
    ...
    extensions: list = [ExtensionHelper, DataCubeHelper, THREDDSHelper]
```

huard: Because of data ingestion. 

For example, the THREDDS extension needs to be passed `THREDDSHelper(data['data']["access_urls"])`

I currently manage this with `model_validator`: 
```
    @model_validator(mode="before")
    @classmethod
    def thredds_helper(cls, data):
        """Instantiate the THREDDSHelper."""
        data["thredds"] = THREDDSHelper(data['data']["access_urls"])
        return data
```
I'm not sure how we'd do that with your proposal without adding some obscure magic. 
What I've now done is automatically detect extensions from the annotation (if it's a Helper subclass). Hope that's ok for now.

huard:
I've added a mechanism to embed the schema into a STAC item schema, that I save in /tmp for now. This was just for testing. We can disable that feature for now. I'll be on travel starting tomorrow, so won't be able to work on this much.

tests/data/cordex6_ncml.json
fmigneault, line 1: Please format the contents to help reading them and detect partial changes.

When trying to format it myself, I found that there are `NaN` that makes it an invalid JSON.

tests/data/cordex6_raw.json
fmigneault, line 1: same as other file

STACpopulator/extensions/base.py
fmigneault: I was actually misinterpreting the question thinking of `extensions` as the `STAC.Item.extensions` (ie: the URI to the applied extensions). I think we should rename the attribute, because that is very confusing. It should be `helpers` to highlight the use of the helpers that have extended capabilities for applying the STAC extensions (and sometimes non-extension attributes). 

Ideally, we should have something like: 
```python
helpers: list[Type[Helper]] = [ExtensionHelper, DataCubeHelper, THREDDSHelper]
```

Because only classes of the helpers are used (not instances), they should be able to receive the item data dynamically for the `apply()` call.

huard: Renamed `_extensions` to `_helpers`. 

I don't see how I could implement your proposal, without hard-coding the `data` ingestion logic into the helpers themselves, which would couple them tightly with the `Loader`, which I thought we should avoid.

fmigneault: Minor nitpick. 

I think the `BaseSTAC` class should define something like: 

```python
class BaseSTAC(abc.ABC):
   @classmethod
   @abc.abstractmethod
   def helpers(cls) -> list[Type[Helper]]:
       raise NotImplementedError
```

This way, any derived class and smart IDE flags right away that `helpers` must be overridden.

I think `helpers` should be used instead of `_helpers` because it is part of the "public" interface of that class, for anyone that derives a new implementation from it.

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,331 @@
...
+
+    model_config = ConfigDict(populate_by_name=True, extra="ignore", arbitrary_types_allowed=True)
+
+    # Helpers are automatically detected by being Helper subclasses
+    _helpers: list[str] = PrivateAttr([])
+
+    @property
+    def uid(self) -> str:
+        """Return a unique ID. When subclassing, use a combination of properties uniquely identifying a dataset."""
+        # TODO: Should this be an abstract method?
```
mishaschwartz, line 157: Yeah I think it probably should.
Especially since the intention is for it to be overridden when this is subclassed.

Also this may cause problems if a subclass forgets to redefine this and then it is called multiple times within the subclass... it will give a different uid each time:

```python
>>> item = BaseSTAC(geometry=None, bbox=[], start_datetime='1999-01-01', end_datetime='1999-01-02')
>>> item.uid
'd33916d8-7bf6-4bf3-9659-e590badc96c6'
>>> item.uid
'b6729ec6-ced6-4076-99fe-e0df321bd163'
```

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,330 @@
...
+                   datacube=DataCubeHelper(data),  # A bit clunky to avoid breaking CMIP6
+                   thredds=THREDDSHelper(data["access_urls"])
+                   )
+
+    @property
+    def uid(self) -> str:
+        """Return a unique ID. When subclassing, use a combination of properties uniquely identifying a dataset."""
+        # TODO: Should this be an abstract method?
+        import uuid
+        return str(uuid.uuid4())
```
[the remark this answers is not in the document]
mishaschwartz: I agree with all this except that I don't think it should be a class method.
When it is overridden in subclasses, this needs access to attributes of the instance (not just the class) to generate a unique id for that specific instance.

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,331 @@
...
+
+    @model_validator(mode="after")
+    def find_helpers(self):
```
mishaschwartz, line 162: this should return `self` or else pydantic raises a UserWarning

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,331 @@
...
+        # Add extensions
+        for ext in self._helpers:
+            getattr(self, ext).apply(item)
+
+        try:
+            item.validate()
+        except STACValidationError as e:
+            raise Exception("Failed to validate STAC item") from e
+
+        return json.loads(json.dumps(item.to_dict()))
```
mishaschwartz, line 189: Why `dumps` and then `loads`? Do we not trust that `to_dict` will be correct?

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,331 @@
...
+    attrs = {"name": name, "schema_uri": schema_uri}
+    return types.new_class(name=cls_name, bases=bases, kwds=None, exec_body=lambda ns: ns.update(attrs))
+
+
+class MetaExtension:
+    name: str
+    schema_uri: str
+
+    def apply(self, properties: dict[str, Any]) -> None:
+        """Applies CMIP6 Extension properties to the extended
```
mishaschwartz, line 211: CMIP6?

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,331 @@
...
+            self._set_property(prop, val)
+
+    @classmethod
+    def get_schema_uri(cls) -> str:
+        """We have already validated the JSON schema."""
+        return cls.schema_uri
+
+    @classmethod
+    def has_extension(cls, obj: S):
+        # FIXME: this override should be removed once an official and versioned schema is released
```
mishaschwartz, line 224: Does this comment apply here? Is this copied from the CMIP6 code?

Do we need to keep this function since the only use of it seems to be commented out in the CMIP6 code anyway.
And if we can get rid of it can we also get rid of `get_schema_uri`?

STACpopulator/implementations/CORDEXCMIP6_Ouranos/add_CORDEX6.py
```diff
@@ -0,0 +1,63 @@
...
+    data_model = Cordex6DataModel
+    item_geometry_model = None  # Unnecessary, but kept for consistency
+
+    def create_stac_item(self, item_name: str, item_data: dict[str, Any]) -> dict[str, Any]:
+        dm = self.data_model.from_data(item_data)
+        return dm.stac_item()
+
+
+
+# TODO: This probably doesn't need to be copied for every implementation, right ?
```
mishaschwartz, line 25: If there's a lot of common arguments we can factor them out into their own function like we did for add_request_options and add_logging_options

STACpopulator/implementations/CORDEXCMIP6_Ouranos/add_CORDEX6.py
```diff
@@ -0,0 +1,63 @@
...
+    parser.add_argument(
+        "--config",
+        type=str,
+        help=(
+            "Override configuration file for the populator. "
+            "By default, uses the adjacent configuration to the implementation class."
+        ),
+    )
+    add_request_options(parser)
+
```
mishaschwartz, line 46: add: `add_logging_options(parser)` as well

STACpopulator/stac_utils.py
```diff
@@ -123,6 +126,37 @@ def numpy_to_python_datatypes(data: MutableMapping[str, Any]) -> MutableMapping[
...
+    """Convert numpy datatypes to python standard datatypes.
+
+    This is useful when validating against a JSON schema that does not recognize an int32 as an integer.
+
+    Parameters
+    ----------
+    data : dict, list, tuple, int, float, np.integer, np.floating, str
+      Object to convert.
+    """
+    import numpy as np
```
mishaschwartz, line 139: import at the top

mishaschwartz (changes requested):

STACpopulator/extensions/base.py
huard: At no point as a developer do you need to modify the `_helpers` attribute. It's filled automatically by `find_helpers`, so I don't think it has to be "public".

STACpopulator/extensions/base.py
```diff
@@ -0,0 +1,331 @@
...
+        The start datetime of the dataset.
+    end_datetime : datetime
+        The end datetime of the dataset.
+    extensions : list[str]
+        Name of the class attributes that point to STAC extension helper classes. Those extension classes should have an `apply` method.
+    """
+    # STAC item properties
+    geometry: AnyGeometry | None
+    bbox: list[float]
+    start_datetime: datetime
```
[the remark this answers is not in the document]
mishaschwartz: I'd like the option to just have a single time. If the data is from a single point in time it is potentially confusing to require a start and end time.

STACpopulator/extensions/base.py
mishaschwartz: But if there's problems with serialization won't `dumps` fail?

Or is pystac using a different (worse?) serialization library than json?

I'm only balking at this because this seems like a hacky solution to something that should be handled properly by the pystac library.... 

Do you remember what values couldn't be serialized? If so, can we add an issue on pystac and add a comment here about it?

tests/data/cordex6_ncml.json
fmigneault, line 1: Test files have some literal `NaN` that are not valid JSON.

STACpopulator/extensions/base.py
fmigneault: I recall this as well. 

I think it was due to some string-like nested objects (`href` or `media_type` in links/assets, not sure, it's been a while...) not getting converted to valid JSON with `to_dict` directly, and those were not nicely handled by the code receiving them (like `requests` not liking the items to POST or similar). 

It's definitely been a while and many `pystac` updates since, so might be worth a revisit.

fmigneault (commented):
Looks good. 

I suggest we do a `0.7.0` bump before merging this PR to integrate all other bugfixes and quality of life changes on their own (including other opened/approved PRs). This should be a new feature version by itself. Also time to move to a proper "MAJOR" bump after, and have future "features" be actual "MINOR" updates.

tests/data/cordex6_ncml.json
mishaschwartz: NaN literals are valid in python json if the `allow_nan` argument is `True` (which it is by default):

```python
>>> import json
>>> json.loads(json.dumps(float("nan")))
nan
```

mishaschwartz:
@fmigneault 

> I suggest we do a 0.7.0 bump before merging this PR to integrate all other bugfixes and quality of life changes on their own (including other opened/approved PRs).

Ok there's been a few more opened since you commented this so why don't we make 0.7.0 a milestone and ensure that the following are included:

- #79 
- #81 OR #82

All others either depend on this or are still waiting for additional work

tests/data/cordex6_ncml.json
fmigneault: If dealing with JSON within JS or python, that would be fine, but those JSON are POSTed to an API using pure JSON which doesn't define NaN.

fmigneault:
@mishaschwartz 
Yes. Let's proceed this way to integrate everything more easily. 
The PR has conflicts however, so I cannot merge it as is.

huard:
Done

STACpopulator/stac_utils.py
```diff
@@ -123,6 +126,35 @@ def numpy_to_python_datatypes(data: MutableMapping[str, Any]) -> MutableMapping[
...
+        if isinstance(data, tuple):
+            return tuple(out)
+        return out
+
+    elif issubclass(type(data), np.integer):
+        return int(data)
+
+    elif issubclass(type(data), np.floating):
+        return float(data)
+
```
mishaschwartz, line 153: You can add an additional case here to handle the `nan` problem (see https://github.com/crim-ca/stac-populator/pull/64#discussion_r1978019325)

```python
elif isinstance(data, float) and np.isnan(data):
    return None # or "NaN" or however you think it's best to represent it
```

huard:
Done

STACpopulator/stac_utils.py
```diff
@@ -149,7 +149,13 @@ def np2py(data):
         return int(data)
 
     elif issubclass(type(data), np.floating):
-        return float(data)
+        data = float(data)
```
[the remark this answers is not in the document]
fmigneault: Does it work if using directly [`np.isfinite`](https://numpy.org/doc/stable/reference/generated/numpy.isfinite.html) to manage all NaN/±inf/whatever-package combinations?

huard:
So I reverted the changes made earlier, as I believe this is not where a fix should live. In any case, the CORDEX extension has no float attribute, so this should not be an issue. I'll write a dummy extension with a float in to see what the STAC item ends up like.

huard:
Found a pydantic config option to dump NaNs and Inf as strings.

fmigneault:
The `"NaN"` string conversion will resolve to valid/pure JSON. However, is that a valid (ie: "usable") value for that `_FillValue` property? I'm not familiar with the CORDEX dataset, but it looks like the "transparent" no-data value used in rasters. Won't the library parsing this JSON raise an error if `"NaN"` is directly passed to it?

huard:
Probably not, but as far as I understand, it's irrelevant, because the `_FillValue` property is not part of the STAC item properties. 
In any case, at no point in the process are we going to evaluate the json content to instantiate a new `xarray.Dataset` object.

fmigneault (approved):
Can be merged after https://github.com/crim-ca/stac-populator/pull/84

The CHANGES.md will have to be adjusted to move the new items under the relevant 0.7.0 section.

PR #65, opened 2024-10-16
Replace name property by simple attribute

### LINKED ISSUE #63: pystac extension not handling well `name` property
The `CMIP6Extension` has a `name` property:

```
    @property
    def name(self) -> SchemaName:
        return get_args(SchemaName)[0]
```
but this is not handled well by pystac.extensions: https://github.com/stac-utils/pystac/blob/ba406cb7991f992ac6f38026aa779adcc6a4f67a/pystac/extensions/base.py#L258

where `getattr(cls, "name")` returns a property object, not the actual name.

huard:
Fix #63

### DIFF
--- a/STACpopulator/extensions/cmip6.py
+++ b/STACpopulator/extensions/cmip6.py
@@ -49,12 +49,11 @@ except OSError as e:
 
 T = TypeVar("T", pystac.Collection, pystac.Item, pystac.Asset, item_assets.AssetDefinition)
 
-SchemaName = Literal["cmip6"]
 # FIXME: below reference (used as ID in the schema itself) should be updated once the extension is officially released
 # SCHEMA_URI: str = "https://stac-extensions.github.io/cmip6/v1.0.0/schema.json"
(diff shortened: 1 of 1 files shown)


mishaschwartz (approved):
Looks good to me. Is there a reason that the prefix and the name were linked before? Is there any value in trying to enforce that the prefix and the name are the same?

huard:
I suspect not having this consistency might break some of the STAC stuff. Note that this consistency is ensured in the cordex_extension branch. I'll be able to port the changes to CMIP6 once we agree on the framework.

fmigneault (commented):
> Is there a reason that the prefix and the name were linked before?

Yes!

Look at how many duplicate `"cmip6"` hard-coded values there are now. 
It's very easy to forget one here and there. The previous approach allowed for anything to reference the same `SchemaName`. Changing it was a single-place edit. 

The `name` class attribute replacement from a property to fix `pystac` interoperability is good, but I think a single "source of truth" should remain with `SchemaName`.

PR #66, opened 2024-10-16
code cleanup

### LINKED ISSUE #62: Problems running the populator
Running the command found in the README
```python STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py \
    "http://localhost:8880/stac/" \
    "https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/testdata/xclim/cmip6/catalog.html" \
    "STACpopulator/implementations/CMIP6_UofT/collection_config.yml"
```
fails with 
```
usage: add_CMIP6.py [-h] [--update] [--mode {full,single}] [--config CONFIG] [--no-verify] [--cert CERT] [--auth-handler {basic,digest,bearer,proxy,cookie}] [--auth-identity AUTH_IDENTITY]
                    stac_host href
add_CMIP6.py: error: unrecognized arguments: STACpopulator/implementations/CMIP6_UofT/collection_config.yml
```

If I try to run this without the collection path, I get:
```
 python STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py     "http://localhost:8880/stac/"     "https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/testdata/xclim/cmip6/catalog.html"
2024-10-11 14:48:27.908713 [INFO] :: PYESSV :: Loading vocabularies from /home/david/.esdoc/pyessv-archive ... please wait
  File "/home/david/src/stac-populator/STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py", line 137, in <module>
    sys.exit(main())
  File "/home/david/src/stac-populator/STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py", line 133, in main
    return runner(ns)
  File "/home/david/src/stac-populator/STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py", line 123, in runner
    ns.stac_host, data_loader, update=ns.update, session=session, config_file=ns.config, log_debug=ns.debug
AttributeError: 'Namespace' object has no attribute 'debug'
```

If I try to run `cli.py` directly, I get
```
  File "/home/david/src/stac-populator/STACpopulator/cli.py", line 4, in <module>
    import logging
  File "/home/david/src/stac-populator/STACpopulator/logging.py", line 3, in <module>
    import logging.config
ModuleNotFoundError: No module named 'logging.config'; 'logging' is not a package
```

mishaschwartz:
- fix incorrect example in README
- move argument parsing for logging options to the implementation code
- fix bug where logging options were being set incorrectly
- rename files to avoid potential naming conflicts with other packages (`logging` and `requests`)

Resolves #62


huard (approved):

STACpopulator/implementations/CMIP6_UofT/add_CMIP6.py
```diff
@@ -130,8 +132,8 @@ def main(*args: str) -> int:
     parser = argparse.ArgumentParser()
     add_parser_args(parser)
     ns = parser.parse_args(args or None)
-    return runner(ns)
-
+    ns.populator = "CMIP6_UofT"
+    return cli.run(ns)
```
fmigneault, line 136: I think this is potentially dangerous if we add other sub-`ArgumentParser` commands than the top-level CLI `run` command which might not have the `populator` attribute. 

Since the `run` function (not to confuse with the `run` subparser command) only does `implementation_modules()[ns.populator].runner(ns)`, there's basically no difference from doing `runner(ns)` (as before), and there's no need to inject the `populator` argument anymore.

mishaschwartz: This is necessary so that the logging options are applied when this file is executed directly from the command line.

> I think this is potentially dangerous if we add other sub-ArgumentParser commands than the top-level CLI run command which might not have the populator attribute.

Agreed, if that happens, then we'd have to modify this code as well. The simpler and more maintainable thing to do would be to remove the option to call populator scripts directly from the command line.

fmigneault: I'm not sure to understand why the logging options cause this? `populator` is only defined by the `run` subparser. What causes `populator` to become an issue if calling the script directly since `run` would be bypassed entirely?

fmigneault: I gave it a try, and no error on my end (beside the unrelated schema validation of course). 

{BE7604F5-C019-4211-A68C-3435813DA442}

mishaschwartz: > I'm not sure to understand why the logging options cause this?

There's no error but the logging options are applied here https://github.com/crim-ca/stac-populator/blob/b353febe2d703a49f9574cbee9feb3feb131108f/STACpopulator/cli.py#L49

So if you don't run it through the CLI functions then the logging options are never applied. We could have added a call to `setup_logging` in every `runner` function as well but that seemed like we're adding boilerplate (which we're actively trying to remove in #64)

fmigneault: OK. But can we at least resolve the `populator` name automatically rather than a hard coded value? I really dislike having different name resolution strategies by different part of the code. If the directory name is the "reference" for the populator name, then everywhere should resolve it the same way.

fmigneault (approved):

PR #73, opened 2025-02-07
fix tests to allow for variable STAC versions in output

mishaschwartz:
Tests currently assume that the STAC version in item and collection JSONs will always be `1.0.0`.
Since we don't pin a specific version of `pystac`, a new version of `pystac` uses STAC version `1.1.0` which now causes tests to fail.

This PR updates tests to allow for item and collection JSONs to have a `stac_version` field that matches the version used by the installed version of `pystac`.

As part of this PR, I also updated the dependency versions to not allow major version upgrades automatically. For example, `pystac` is now pinned to version `~=1.12` which will mean is will not install version 2+. This should avoid similar breaking changes caused by dependency updates in the future.

### DIFF
--- a/tests/test_standalone_stac_item.py
+++ b/tests/test_standalone_stac_item.py
@@ -6,6 +6,7 @@ import tempfile
 from urllib.parse import quote
 
 import xncml
+from pystac.version import get_stac_version
 
 from STACpopulator.extensions.cmip6 import CMIP6Helper
 from STACpopulator.extensions.thredds import THREDDSHelper, THREDDSExtension
@@ -50,6 +51,7 @@ def test_standalone_stac_item_thredds_ncml():
     ref_file = os.path.join(CUR_DIR, "data/stac_item_testdata_xclim_cmip6_ncml.json")
     with open(ref_file, mode="r", encoding="utf-8") as ff:
         reference = json.load(ff)
+        reference["stac_version"] = get_stac_version()
 
     assert stac_item.to_dict() == reference
 
@@ -86,5 +88,6 @@ def test_cmip6_stac_thredds_catalog_parsing():
     ref_file = os.path.join(CUR_DIR, "data/stac_collection_testdata_xclim_cmip6_catalog.json")
     with open(ref_file, mode="r", encoding="utf-8") as ff:
         reference = json.load(ff)
+        reference["stac_version"] = get_stac_version()
 
     assert result == reference
--- a/tests/test_standalone_stac_item.py
+++ b/tests/test_standalone_stac_item.py
@@ -1,4 +1,5 @@
 import json
+import pystac
 import pytest
 import requests
 import os
@@ -6,7 +7,6 @@ import tempfile
 from urllib.parse import quote
 
 import xncml
-from pystac.version import get_stac_version
 
 from STACpopulator.extensions.cmip6 import CMIP6Helper
 from STACpopulator.extensions.thredds import THREDDSHelper, THREDDSExtension
@@ -50,8 +50,7 @@ def test_standalone_stac_item_thredds_ncml():
 
     ref_file = os.path.join(CUR_DIR, "data/stac_item_testdata_xclim_cmip6_ncml.json")
     with open(ref_file, mode="r", encoding="utf-8") as ff:
-        reference = json.load(ff)
-        reference["stac_version"] = get_stac_version()
+        reference = pystac.Item.from_dict(json.load(ff)).to_dict()
 
     assert stac_item.to_dict() == reference
 
@@ -87,7 +86,6 @@ def test_cmip6_stac_thredds_catalog_parsing():
 
     ref_file = os.path.join(CUR_DIR, "data/stac_collection_testdata_xclim_cmip6_catalog.json")
     with open(ref_file, mode="r", encoding="utf-8") as ff:
-        reference = json.load(ff)
-        reference["stac_version"] = get_stac_version()
+        reference = pystac.Collection.from_dict(json.load(ff)).to_dict()
 
     assert result == reference
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,7 +20,8 @@
 * move argument parsing for logging options to the implementation code
 * fix bug where logging options were being set incorrectly
 * rename files to avoid potential naming conflicts with other packages (`logging` and `requests`)
-
+* enforce versions for dependencies so that new installs won't fail unexpectedly
+* update tests to allow for a variable `stac_version` field in STAC item and collections 
 
 ## [0.6.0](https://github.com/crim-ca/stac-populator/tree/0.6.0) (2024-02-22)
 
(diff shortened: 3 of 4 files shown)


tests/test_standalone_stac_item.py
```diff
@@ -86,5 +88,6 @@ def test_cmip6_stac_thredds_catalog_parsing():
     ref_file = os.path.join(CUR_DIR, "data/stac_collection_testdata_xclim_cmip6_catalog.json")
     with open(ref_file, mode="r", encoding="utf-8") as ff:
         reference = json.load(ff)
+        reference["stac_version"] = get_stac_version()
```
fmigneault, line 91: Seems problematic IMO. 

There are different properties (eg: new `bands`) in STAC 1.1 for which `stac_version` should not be blindly "overridden", but actually migrated. 

If there was an error when using `pystac` within `stac-populator`, I'd argue this is probably a mishandling somewhere which should be addressed. I find this test properly early-detected an issue that must be investigated further, since `pystac` is supposed to support both versions, and do appropriate migrations as needed.

Are we not simply ignoring a potential issue if the catalog targeted by `stac-populator` still uses STAC 1.0? We cannot just push a STAC 1.1 to it, and the populator code should adapt to this (if needed).

fmigneault (commented):
Without some kind of dependapot checking packages regularly, I'm expecting maintenance creep and vulnerabilities in the long run.

For example, `requests` and `lxml` are regularly updated to resolve security issues. I somewhat like the fact that those are auto-bumped to latest version. The "pinned" versions, if needed, are available in the docker.

Note also that limiting MAJOR is not really a fix, just as in the case of `pystac`. It is a seemingly "MINOR" update that broke the tests by changing STAC version entirely.

tests/test_standalone_stac_item.py
fmigneault: See https://github.com/stac-utils/pystac/blob/main/CHANGELOG.md#v1121

Maybe we should have `migrate=False` explicitly?

mishaschwartz: > Are we not simply ignoring a potential issue if the catalog targeted by stac-populator still uses STAC 1.0? We cannot just push a STAC 1.1 to it, and the populator code should adapt to this (if needed).

I don't think that is what this is testing. This does not try to upload any data to an existing STAC server so it doesn't have any information from a server to know what version it is running.

We should definitely add a test that checks for that but that should be additional to this one. 

> There are different properties (eg: new bands) in STAC 1.1 for which stac_version should not be blindly "overridden", but actually migrated.

I'll update this test to ensure that any migrations beyond just the "stac_version" between versions are being tested as well.

mishaschwartz:
> Without some kind of dependapot checking packages regularly, I'm expecting maintenance creep and vulnerabilities in the long run.

Ok let's add a dependabot

> For example, requests and lxml are regularly updated to resolve security issues. I somewhat like the fact that those are auto-bumped to latest version. The "pinned" versions, if needed, are available in the docker.

We're not just developing a docker image though, this is also usable as a package and dependency versions should be managed properly so they don't break things for other users.

> Note also that limiting MAJOR is not really a fix, just as in the case of pystac. It is a seemingly "MINOR" update that broke the tests by changing STAC version entirely.

Ok so we'll keep pystac more limited. Good to know.

mishaschwartz:
> > Without some kind of dependapot checking packages regularly, I'm expecting maintenance creep and vulnerabilities in the long run.
> 
> Ok let's add a dependabot

FYI I don't have sufficient permissions to enable dependabot for this repo. If you have permissions @huard or @fmigneault please feel free to set it up.

tests/test_standalone_stac_item.py
fmigneault: What I mean is that the populator code should work with either 1.0 or 1.1 STAC documents. 

If there is any dependency (`pystac` here) that suddenly breaks because 1.0 isn't parsed as expected, or results in a different output, the STAC Item that *would eventually be pushed* to the referenced catalog could fail during that step, since that catalog might not yet support more recent versions.

Even if the test indeed doesn't upload anything here, we were "lucky" to actually detect this here. `pystac` should not impose its own newer version if the STAC document already indicated one (as it does here), since it might have a legitimate reason to do so, just like using 1.1 explicitly most probably means that item relies on the new `bands` property.

fmigneault (changes requested):
waiting on the `stac_version` adjustment, dependabot configured

tests/test_standalone_stac_item.py
mishaschwartz: The STAC document doesn't indicate anything here. It is used as a reference in the test to check that the pystac item matches some hard-coded json. At no point is that reference json used as an input to pystac.

If we need a new test or new code to ensure that pystac respects existing versions then great! Good idea! Let's add that in. Let's make an issue for it. Let's add a new test in a different PR that checks for that.

This pull request is trying to fix existing tests so that our test suite continues to work as it did before. We cannot merge any existing PRs without the test suite passing.

If you will not approve this PR until the test you want has been added then fine. Either me or you or @huard can add that test when we have the time to dedicate to this and all the other PRs will wait in the meantime.

fmigneault: Fair point. Then, it raises another issue, how to indicate to `pystac` that it should generate `stac_version: 1.0.0` if only this is supported by the remote catalog? I didn't dig into the code, but I would assume this can be specified by parameter somewhere, which this test could set explicitly. If not, `pystac` must be pinned to a maximum version until this is made possible.

IMO, this test raises a valid concern. Running the populator with an older version vs after these package updates are applied yield different results. Therefore, we cannot rely on the contents it generates. Simply patching the hardcoded contents of test does not address the reproducibility problem.

mishaschwartz: From the [pystac docs](https://pystac.readthedocs.io/en/stable/concepts.html#stac-spec-version-support): 

> PySTAC supports STAC Spec 1.1.0 and will automatically update any catalogs to this version

They suggest:

> To work with older versions of the STAC Spec, please use an older version of PySTAC

To use version 1.0.0 of the STAC spec, we'd have to use version 1.0 of PySTAC. Obviously this means that you will need separate installations of pystac if you want to be able to support multiple versions of the STAC spec.

You can [set the version number](https://pystac.readthedocs.io/en/stable/api/pystac.html#pystac.set_stac_version) in the STAC item/collection but this does not change the content of the STAC item/collection itself to correspond to the version. They say:

> Setting the STAC version to something besides the default version will not effect the format of STAC read or written; it will only override the stac_version property of the objects being written. Setting this incorrectly can produce invalid STAC.

The current suggestion that comes up in the discussions around this is to use https://github.com/stac-utils/stac-migrate to migrate existing catalogs that use older versions.

### Summary

Pystac does not support creating catalogs, collections or items that conform with older versions of the stac-spec

If we want this we will need a different strategy entirely.

mishaschwartz: @fmigneault 

I have a proposed solution for this issue in #79. 

Can we please merge this PR so that the initial issue of failing tests is resolved and then the secondary issue can be discussed in the other PR if needed.

fmigneault: I'm not sure the solution is quite right yet.
See my comment directly on the PR.

fmigneault: Oh, I just saw your https://github.com/crim-ca/stac-populator/pull/73#discussion_r1962385764 after the PR review. 

Seems like the tooling is a limiting factor, so I guess we don't have much choice to move forward... 

Is the currently packaged STAC-FastAPI able to support 1.1 seemlessly? If so, then I think pinning `pystac` in #79 would be a good enough tradeoff.

mishaschwartz: @fmigneault 

> Is the currently packaged STAC-FastAPI able to support 1.1 seemlessly? 

Yes, probably since version 1.1 is compatible with previous versions: https://github.com/stac-utils/pystac/issues/1375#issuecomment-2621537397

Can we continue this discussion in #79 and if you're happy with the changes made in the PR to simply get the tests working again can you please approve this PR.

I can't merge this PR without your approval and I can't merge any other PR that is currently open while the tests are failing.

fmigneault: OK. Let's do this to unblock the PRs, and we address 1.1 separately after.

fmigneault (approved):

PR #77, opened 2025-02-12
add dependabot

fmigneault:


fmigneault:
This is what I did. Seems the auto-generated file from that procedure is missing some config params.

fmigneault:
Now running: https://github.com/crim-ca/stac-populator/network/updates/963398171
Let me know if you think more configuration should be added.

mishaschwartz:
Oh I see, you pushed the change to master directly?

It seems to be working now. Thanks

PR #81, opened 2025-02-27
fix inconsistent update defaults

### LINKED ISSUE #80: Mixed `update` flag defaults leading to confusion
As shown below, the default parameters between the methods and the CLI are opposite (omitting the update flag omits the parameter, which ends up setting `update=True`). They are also inconsistent between the implementations.

Those should be aligned to avoid user confusion when invoking the tool. 

https://github.com/search?q=repo%3Acrim-ca%2Fstac-populator%20Optional%5Bbool%5D%20%3D%20True&type=code
Image

https://github.com/search?q=repo%3Acrim-ca%2Fstac-populator%20--update&type=code
Image

mishaschwartz:
resolves #80 
closes #82

I'm a little confused now what is supposed to happen if  the `--update` flag is not used. Is it supposed to create items/collections/assets only if they don't exist? Is it supposed to report the status of something?
All of the tests currently assume that `--update` is set and I think that we should add some tests for the opposite case but I don't understand what the intended outcome is... :man_shrugging:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,7 @@
 * fix bug where populator scripts could not be called directly from the command line
 * enforce versions for dependencies so that new installs won't fail unexpectedly
 * update tests to allow for a variable `stac_version` field in STAC item and collections 
+* fix inconsistent defaults for parameters that update stac items and collections
 
 
 ## [0.6.0](https://github.com/crim-ca/stac-populator/tree/0.6.0) (2024-02-22)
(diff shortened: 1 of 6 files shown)


huard (commented):
I'm confused as well. I think update and add_if_missing are two different concepts. My superficial understanding is that update will update the content of an existing item, while add_if_missing only deals with the schema URI.

mishaschwartz:
#82 is an alternative PR that doesn't touch `add_if_missing` and might be more what @fmigneault had in mind. Let me know

PR #87, opened 2025-05-15
Add ability to export data from a STAC catalog or API to files on disk.

mishaschwartz:
This add the `export` CLI option and associated code which exports STAC data from an existing API or catalog and stores the JSON files on disk. 

This is related to https://github.com/bird-house/birdhouse-deploy/issues/536

### DIFF
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -38,7 +38,7 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     export_parser.add_argument("stac_host", help="STAC API URL")
     export_parser.add_argument("directory", type=str, help="Path to a directory to write STAC catalog contents.")
     export_parser.add_argument(
-        "-c", "--continue", action="store_true", dest="continue_", help="Continue a partial download."
+        "-r", "--resume", action="store_true", help="Resume a partial download."
     )
     add_request_options(export_parser)
     add_logging_options(export_parser)
@@ -70,7 +70,7 @@ def run(ns: argparse.Namespace) -> int:
     else:
         with requests.Session() as session:
             apply_request_options(session, ns)
-            return export_catalog(ns.directory, ns.stac_host, session, ns.continue_) or 0
+            return export_catalog(ns.directory, ns.stac_host, session, ns.resume) or 0
 
 
 def main(*args: str) -> int:
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -6,9 +6,13 @@ import sys
 import warnings
 from types import ModuleType
 
+import requests
+
 from STACpopulator import __version__, implementations
 from STACpopulator.exceptions import STACPopulatorError
-from STACpopulator.log import setup_logging
+from STACpopulator.export import export_catalog
+from STACpopulator.log import add_logging_options, setup_logging
+from STACpopulator.request_utils import add_request_options, apply_request_options
 
 
 def add_parser_args(parser: argparse.ArgumentParser) -> None:
@@ -30,6 +34,14 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     for implementation_module_name, module in implementation_modules().items():
         implementation_parser = populators_subparser.add_parser(implementation_module_name)
         module.add_parser_args(implementation_parser)
+    export_parser = commands_subparser.add_parser("export", description="Export a STAC catalog to JSON files on disk.")
+    export_parser.add_argument("stac_host", help="STAC API URL")
+    export_parser.add_argument("directory", type=str, help="Path to a directory to write STAC catalog contents.")
+    export_parser.add_argument(
+        "-c", "--continue", action="store_true", dest="continue_", help="Continue a partial download."
+    )
+    add_request_options(export_parser)
+    add_logging_options(export_parser)
 
 
 @functools.cache
@@ -52,9 +64,13 @@ def implementation_modules() -> dict[str, ModuleType]:
 
 def run(ns: argparse.Namespace) -> int:
     """Run a given implementation given the arguments passed on the command line."""
+    setup_logging(ns.log_file, ns.debug or logging.INFO)
     if ns.command == "run":
-        setup_logging(ns.log_file, ns.debug or logging.INFO)
         return implementation_modules()[ns.populator].runner(ns) or 0
+    else:
+        with requests.Session() as session:
+            apply_request_options(session, ns)
+            return export_catalog(ns.directory, ns.stac_host, session, ns.continue_) or 0
 
 
 def main(*args: str) -> int:
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -37,9 +37,7 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     export_parser = commands_subparser.add_parser("export", description="Export a STAC catalog to JSON files on disk.")
     export_parser.add_argument("stac_host", help="STAC API URL")
     export_parser.add_argument("directory", type=str, help="Path to a directory to write STAC catalog contents.")
-    export_parser.add_argument(
-        "-r", "--resume", action="store_true", help="Resume a partial download."
-    )
+    export_parser.add_argument("-r", "--resume", action="store_true", help="Resume a partial download.")
     add_request_options(export_parser)
     add_logging_options(export_parser)
 
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -38,6 +38,11 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     export_parser.add_argument("stac_host", help="STAC API URL")
     export_parser.add_argument("directory", type=str, help="Path to a directory to write STAC catalog contents.")
     export_parser.add_argument("-r", "--resume", action="store_true", help="Resume a partial download.")
+    export_parser.add_argument(
+        "--ignore-duplicate-ids",
+        action="store_true",
+        help="Do not raise an error if STAC items with the same ids are found in a collection.",
+    )
     add_request_options(export_parser)
     add_logging_options(export_parser)
 
@@ -68,7 +73,7 @@ def run(ns: argparse.Namespace) -> int:
     else:
         with requests.Session() as session:
             apply_request_options(session, ns)
-            return export_catalog(ns.directory, ns.stac_host, session, ns.resume) or 0
+            return export_catalog(ns.directory, ns.stac_host, session, ns.resume, ns.ignore_duplicate_ids) or 0
 
 
 def main(*args: str) -> int:
--- a/STACpopulator/export.py
+++ b/STACpopulator/export.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import pathlib
+import time
 from typing import Iterable, cast
 
 import pystac
@@ -12,6 +13,10 @@ import requests
 from pystac_client.stac_api_io import StacApiIO
 
 LOGGER = logging.getLogger(__name__)
+
+
+class DuplicateIDError(Exception):
+    """Duplicate ID Error."""
 
 
 class Client(pystac_client.Client):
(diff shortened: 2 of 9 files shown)


huard (approved):
Just gave a sperficial look, but agree with the intent and the interface is clear.

STACpopulator/cli.py
```diff
@@ -30,6 +34,14 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     for implementation_module_name, module in implementation_modules().items():
         implementation_parser = populators_subparser.add_parser(implementation_module_name)
         module.add_parser_args(implementation_parser)
+    export_parser = commands_subparser.add_parser("export", description="Export a STAC catalog to JSON files on disk.")
+    export_parser.add_argument("stac_host", help="STAC API URL")
+    export_parser.add_argument("directory", type=str, help="Path to a directory to write STAC catalog contents.")
+    export_parser.add_argument(
+        "-c", "--continue", action="store_true", dest="continue_", help="Continue a partial download."
```
fmigneault, line 41: `resume` to avoid the `_` ?

STACpopulator/export.py
```diff
@@ -0,0 +1,104 @@
...
+            self.remove_conforms_to("ITEM_SEARCH")
+            yield from super().get_items(*ids, recursive=recursive)
+
+
+# Ensure that nested catalogs use our updated Client class
+pystac_client.client.Client = Client
+
+
+def _export_catalog(
+    client: Client | pystac_client.CollectionClient, directory: os.PathLike, continue_: bool = False
```
fmigneault, line 78: Seems like `directory: pathlib.Path` should be used since `/=` won't work on a literal string (e.g.: allowed by `os.PathLike`).

STACpopulator/export.py
```diff
@@ -0,0 +1,104 @@
...
+) -> None:
+    directory /= client.id
+    if not continue_ and directory.exists():
+        n_duplicates = sum(1 for _ in directory.parent.glob(f"{client.id}*/"))
```
fmigneault, line 82: Why use `sum(1...)` instead of `len(glob())` ?

STACpopulator/export.py
```diff
@@ -0,0 +1,104 @@
...
+        LOGGER.warning(
+            "Catalog or collection with ID %s already exists in this catalog. IDs should be unique!", client.id
+        )
+    directory.mkdir(exist_ok=True, parents=True)
+    file_name = "catalog.json" if isinstance(client, Client) else "collection.json"
+    with open(directory / file_name, "w") as f:
+        json.dump(client.to_dict(transform_hrefs=False), f)
+    for item in client.get_items(recursive=False):
+        with open(directory / f"item-{item.id}.json", "w") as f:
+            json.dump(item.to_dict(transform_hrefs=False), f)
```
fmigneault, line 93: Consider explicit `mode="w", encoding="utf-8"`.

STACpopulator/export.py
```diff
@@ -0,0 +1,104 @@
...
+def _export_catalog(
+    client: Client | pystac_client.CollectionClient, directory: os.PathLike, continue_: bool = False
+) -> None:
+    directory /= client.id
+    if not continue_ and directory.exists():
+        n_duplicates = sum(1 for _ in directory.parent.glob(f"{client.id}*/"))
+        directory = directory.parent / f"{client.id}-duplicate-id-{n_duplicates}"
+        LOGGER.warning(
+            "Catalog or collection with ID %s already exists in this catalog. IDs should be unique!", client.id
+        )
```
fmigneault, line 86: Could use the same check as filename below to distinguish between catalog/collection and provide clearer indication (might not be obvious if UUIDs are used).

STACpopulator/export.py
```diff
@@ -0,0 +1,104 @@
...
+    with open(directory / file_name, "w") as f:
+        json.dump(client.to_dict(transform_hrefs=False), f)
+    for item in client.get_items(recursive=False):
+        with open(directory / f"item-{item.id}.json", "w") as f:
+            json.dump(item.to_dict(transform_hrefs=False), f)
+    for child in client.get_children():
+        _export_catalog(child, directory, continue_=continue_)
+
+
+def export_catalog(directory: os.PathLike, stac_host: str, session: requests.Session, continue_: bool = False) -> int:
```
fmigneault, line 98: Not returning `int`. Missing or copy-paste mistake?

tests/cassettes/test_export/test_export_api.yaml
```diff
@@ -0,0 +1,2940 @@
...
+    body:
+      string: '{"id": "c604ffb6d610adbb9a6b4787db7b8fd7", "type": "Collection", "links":
+        [{"rel": "items", "type": "application/geo+json", "href": "https://hirondelle.crim.ca/stac/collections/c604ffb6d610adbb9a6b4787db7b8fd7/items"},
+        {"rel": "parent", "type": "application/json", "href": "https://hirondelle.crim.ca/stac/"},
+        {"rel": "root", "type": "application/json", "href": "https://hirondelle.crim.ca/stac/"},
+        {"rel": "self", "type": "application/json", "href": "https://hirondelle.crim.ca/stac/collections/c604ffb6d610adbb9a6b4787db7b8fd7"},
+        {"rel": "items", "href": "http://stac:8000/stac/collections/c604ffb6d610adbb9a6b4787db7b8fd7/items",
+        "type": "application/geo+json"}, {"rel": "items", "href": "http://stac:8000/stac/collections/c604ffb6d610adbb9a6b4787db7b8fd7/items",
+        "type": "application/geo+json"}, {"rel": "items", "href": "http://stac:8000/stac/collections/c604ffb6d610adbb9a6b4787db7b8fd7/items",
+        "type": "application/geo+json"}, {"rel": "items", "href": "http://stac:8000/stac/collections/c604ffb6d610adbb9a6b4787db7b8fd7/items",
```
fmigneault, line 2801: Still not patched. 😅 
The invalid URLs do not cause issues with the client?

tests/cassettes/test_export/test_export_api.yaml
```diff
@@ -0,0 +1,2940 @@
...
+      - gzip, deflate
+      Connection:
+      - keep-alive
+      User-Agent:
+      - python-requests/2.32.3
+    method: GET
+    uri: https://hirondelle.crim.ca/stac
+  response:
+    body:
+      string: '{"type": "Catalog", "id": "stac-fastapi", "title": "Data Analytics
```
fmigneault, line 17: Not related to this PR, but I guess we will have to address that in birdhouse-deploy to avoid conflicting IDs between our instances...

STACpopulator/export.py
mishaschwartz: `glob` returns a map object which doesn't have a length. So you have to iterate over it anyway to get the length. You could do something like: `len(list(glob(...)))` but then you have a list in memory of every object. If you have to iterate over the map, why not just get the information you need (a count)?

STACpopulator/export.py
```diff
@@ -0,0 +1,104 @@
...
+pystac_client.client.Client = Client
+
+
+def _export_catalog(
+    client: Client | pystac_client.CollectionClient, directory: os.PathLike, continue_: bool = False
+) -> None:
+    directory /= client.id
+    if not continue_ and directory.exists():
+        n_duplicates = sum(1 for _ in directory.parent.glob(f"{client.id}*/"))
+        directory = directory.parent / f"{client.id}-duplicate-id-{n_duplicates}"
```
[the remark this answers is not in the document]
mishaschwartz: I chose to do it this way because there are several catalogs out in the wild that don't have unique ids. For example:

https://asc-jupiter.s3.us-west-2.amazonaws.com/europa/galileo_voyager/controlled_mosaics/catalog.json and https://asc-jupiter.s3.us-west-2.amazonaws.com/europa/galileo_voyager/controlled_images/catalog.json both have the id `"usgs_galileo_controlled_images_catalog"` and both are subcatalogs of https://asc-jupiter.s3.us-west-2.amazonaws.com/europa/galileo_voyager/catalog.json .

The spec [requires that ids be unique for collections](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#collection-fields) but for nested catalogs it just [recommends](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#id) that they be unique. So I don't think that we should be raising an error for non-unique catalog ids which are still technically spec-compliant. We could stop for non-unique collections though...

> Similarly, there could be some --override flag to allow recreating the files when resume is specified, whereas omitting it would (more quickly) skip over duplicate collections/item IDs?

I also considered this but... it doesn't really add much time saving because you still have to get all of the collection/item data in order to determine the id. If the id was guaranteed to be in the item URL and if we didn't use the `/search` endpoint to find items then yes this would be more efficient, but with the way that STAC is set up this doesn't really buy us anything.

tests/cassettes/test_export/test_export_catalog_nested.yaml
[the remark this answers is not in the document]
mishaschwartz: No they're all unrelated to login/security (see: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html):

- x-amz-id-2: used for troubleshooting
- x-amz-request-id: unique request id 
- x-amz-meta-mtime: file mtime 

They're not needed for the tests so I can remove them (actually none of the headers are needed really).

tests/cassettes/test_export/test_export_api.yaml
mishaschwartz: No because the client either looks for links with `"rel" = "item"` (not `"items"`) or it uses the `/search` endpoint.

The [`collections/{id}/items`](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0/ogcapi-features#collection-items-collectionscollectioniditems) endpoint isn't used directly by `pystac-client` when looking for items for some reason.

tests/cassettes/test_export/test_export_api.yaml
mishaschwartz: Yeah good point.

STACpopulator/export.py
fmigneault: I see, my bad. I was testing using `glob.glob` which handles `len` rather than pathlib's `glob`.

STACpopulator/export.py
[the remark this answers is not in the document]
fmigneault: I think the handling of duplicate ID catalogs is a valid use case, as illustrated by your example. No problem in supporting this use case.

It's just that, if the client/user expects to read these IDs to locate the catalog directories exported on disk afterward, it won't resolve by name. Unless these IDs are an actual error in these remote STAC catalogs, they should technically be interpreted similarly to a symlink/redirect on disk. They *should* be the "same content". 

In this kind of situation where there are voluntary duplicate-IDs-but-differing-contents to export, I would rather have this behavior "forced" by the user with an explicit flag, since it goes against what STAC specification requires. This action should be deliberate, not a side effect. The default behavior would be to warn the user that "duplicate STAC catalogs" are skipped.

>  it doesn't really add much time saving

Really? I would expect skipping very large JSON load/dump + their respective HTTP requests to save a lot of time if we can preemptively skip based on an existing item ID and the same item IDs found in the collection's listing requests. Why use `/search`, since we want to export everything?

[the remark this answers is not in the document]
mishaschwartz: > they should technically be interpreted similarly to a symlink/redirect on disk

Not really, since they are accessed at different URLs they just happen to have the same id. My guess is that they have the same id due to user error, not anything intentional.

> The default behavior would be to warn the user that "duplicate STAC catalogs" are skipped.

The problem is that they're not duplicate because the content is different and they're accessed at a different URLs. It's just that the IDs are messed up.

> Really? I would expect skipping

If we stop looking for anything once we have a collection that's already downloaded then sure. But if we want to support a partially downloaded collection then we need to inspect every sub-collection and item in order to determine if we already have downloaded one of them. In that case it saves a JSON load but not the HTTP request (which is what takes the most time).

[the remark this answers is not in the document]
fmigneault: > My guess is that they have the same id due to user error, not anything intentional.

I believe this is the case also, hence why I want the operation to be deliberate by the user doing the export. As mentioned, it *should* be the same according to STAC, but this is definitely not the case in the wild. I want users to have the option to ignore it since we also need it, but they must be made aware of this potentially breaking consideration. Not flagging the issue by default is prone to errors. For instance, the order in which URLs are returned would change which one gets the duplicate suffix or not. If an import script ever relies on the directory name, that would make inconsistent results. 

> In that case it saves a JSON load but not the HTTP request (which is what takes the most time).

Fair point. 
I guess it depends on the STAC API implementation. 

Technically, STAC spec allows listing all `item` links directly under `/collections/{id}`, which could speed up parsing by skipping the load+request of all STAC Items and their respective properties altogether. However, there seems to be a growing trend toward listing all contents directly in the paged `/collections/{id}/items` response, so not much optimization really applies anymore...

OK to ignore this concern for now. Not blocking for the PR.

tests/cassettes/test_export/test_export_api.yaml
fmigneault: That explains it! I really was under the impression that the `ItemCollection` iterator was simply using that endpoint. Seems unnecessary to do a catalog wide search when listing all collections and their items one by one. Doesn't that cause unnecessary spatio-temporal filters/compares that inevitably returns everything?

STACpopulator/export.py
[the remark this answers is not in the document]
mishaschwartz: OK done. I added an explicit `--ignore-duplicate-ids` flag. A user must specify this flag or else an error is raised when a duplicate id is encountered.

tests/cassettes/test_export/test_export_api.yaml
mishaschwartz: I think that depends on what the API in the backend does.

In my opinion, pystac-client makes some odd choices. I think there are definitely places that it could be improved.

fmigneault: Yeah. Seems to be using the `Catalog` client's `ItemSearch` by default, since it's the most generic way that allows finding items across collections and even the collectionless items.

To take advantage of iterative optimizations mentioned in other comments that could resume/skip processed collections/items by ID, you would have to explicitly use `Catalog.get_collections` to yield `CollectionClient` and call `CollectionClient.get_items(recursive=True)`. Speed up is probably only noticeable if resuming a download midway of very large collections, or to retrieve only the few newly recently added items to the entire catalog without reloading all other JSONs.

STACpopulator/export.py
```diff
@@ -74,32 +79,85 @@ def get_items(self, *ids: str, recursive: bool | None = None) -> Iterable[pystac
...
+    whether a file contains a non-unique STAC object (i.e. this does not attempt to lock the file).
+    """
+    if file.exists():
+        if start_time < file.stat().st_mtime:
+            msg = (
+                f"{data['type']} with ID {data['id']} already exists in directory {file.parent}. IDs should be unique!"
+            )
+            if ignore_duplicate_ids:
+                LOGGER.warning(msg)
+                n_duplicates = sum(1 for _ in file.parent.glob(f"{file.name} [0-9]"))
```
fmigneault, line 108: What happens when/if `<file> 9` is reached? If that happens, won't this be capped at `n_duplicates = 10` since the `<file> 10` would not match? The glob doesn't need a `[0-9]+` or something similar?

Also, there is technically a risk of `<file> [0-9]` to be a "valid" ID already, since the spec does not technically constrain `id` to non-spaced slugs. Is there another character or strategy to consider addressing this, or we consider it an acceptable edge case? Either way, a comment might be relevant if some specific logic/reason applies for a specifically selected pattern.

mishaschwartz: Good point about duplicates > 9. I've now handled that... globbing doesn't support `+` but we can get the same results with `[0-9]*` since all values after the initial digit will always be digits as well.

> Also, there is technically a risk of <file> [0-9] to be a "valid" ID already

No there isn't because the file name includes the `.json` extension and the digit that indicates that there's a duplicate goes after the file extension. 

This is intentional to avoid conflicts and it also makes it clear to the user who may be looking at the files in a folder that there are clearly duplicates. 

For example, if we already have a catalog with id `catalogA` then we get a file like, `catalogA/catalog.json`. Then if we have another catalog with the same id then we get `'catalogA/catalog.json 1'`. If we later have a catalog with id `'catalog.json 1'` for some strange reason then we would get the file `'catalog.json 1/catalog.json'`.

mishaschwartz: Sure! done

PR #89, opened 2025-05-31
don't log to a file by default

mishaschwartz:
By default a log file will be created for every run. This means that after a while your current directory gets filled up with log files. 

The only way to prevent this was to log to `/dev/null` which is fine but doesn't feel very user friendly.

I would prefer to log to stderr only by default and to give the option to log to a file only when explicitly asked.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
 * Add an `xscen` extension demonstrating how to add properties to a STAC Item.
 * Fix mismatch between CMIP6 schema URI given to `pystac` and the actual schema URI
 * Add ability to export data from a STAC catalog or API to files on disk.
+* Fix code that raised warnings from dependencies.
 * Log to stderr only by default and to a file only if requested.
 
 ## [0.7.0](https://github.com/crim-ca/stac-populator/tree/0.7.0) (2025-03-07)
--- a/STACpopulator/extensions/cmip6.py
+++ b/STACpopulator/extensions/cmip6.py
@@ -21,7 +21,7 @@ from pydantic import (
     BaseModel,
     ConfigDict,
     Field,
-    FieldValidationInfo,
+    ValidationInfo,
     field_validator,
 )
 from pydantic.fields import FieldInfo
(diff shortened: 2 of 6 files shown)


STACpopulator/log.py
```diff
@@ -98,6 +101,12 @@ def filter(self, record: logging.LogRecord) -> bool | logging.LogRecord:
         return record.levelno <= logging.INFO
 
 
+file_handler = {
+    "class": "logging.FileHandler",
+    "level": "INFO",
```
fmigneault, line 106: Would it make sense to reuse the same level as `stderr`, still defined with `DEBUG` in `logconfig`? I think if there's no `--log-level` override, they should at least "align" in what they log.

STACpopulator/log.py
```diff
@@ -144,6 +147,5 @@ def add_logging_options(parser: argparse.ArgumentParser) -> None:
     parser.add_argument("--debug", action="store_const", const=logging.DEBUG, help="set logger level to debug")
     parser.add_argument(
         "--log-file",
-        default=f"stac_populator_log_{dt.datetime.now(dt.timezone.utc).isoformat() + 'Z'}.jsonl",
```
fmigneault, line 147: This line reminded me why `INFO` was used for `log-file` while `DEBUG` for `stderr`. The idea was that `log.info` + `extra_args` were used only for outputting STAC Items and relevant metadata, whereas `stderr` includes any other relevant debugging/step updates.

Now that `--log-level` would be used in the more "traditional" way of simply logging anything to file, maybe it is worth having an option to log only STAC Items metadata as `.jsonl` separately?

STACpopulator/log.py
```diff
@@ -98,6 +101,12 @@ def filter(self, record: logging.LogRecord) -> bool | logging.LogRecord:
         return record.levelno <= logging.INFO
 
 
+file_handler = {
+    "class": "logging.FileHandler",
+    "level": "INFO",
+    "formatter": "json",
```
fmigneault, line 107: Given that no `.jsonl` file name is used by default anymore, I think users will get an unexpected result when their logs are not traditional messages, but per-line JSON. Maybe this should be clarified in the option's description, or with a separate option (see other comment)?

fmigneault (commented):
Very reasonable request. Few usability comments.

STACpopulator/log.py
```diff
@@ -29,23 +29,24 @@
...
-    config = logconfig
-    if logfname is not None:
-        file_handler["filename"] = logfname
-        config["handlers"]["file"] = file_handler
+try:
+    LOG_LEVEL_NAMES = tuple(logging.getLevelNamesMapping())
+except AttributeError:
+    # prior to python 3.11, the logging module did not have a
+    # good way to get level names. Fall back to constants
+    LOG_LEVEL_NAMES = ("CRITICAL", "FATAL", "ERROR", "WARN", "WARNING", "INFO", "DEBUG", "NOTSET")
```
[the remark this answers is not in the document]
fmigneault: Fair enough. We can leave as is since the result is the same.

PR #99, opened 2025-06-25
Fix cordex

huard:
Couple of bug fixes:
- Lingering `log_debug` option in CORDEX CLI runner
- Fix regression introduced in 0.8.0 in the THREDDSLoader iterator. Suggests a gap in the test suite. 
- Add attributes to the CORDEX ID, otherwise different files will get catalogued under the same ID. 
- Update CORDEX test data to latest version

With those, I'm able to create STAC items for our CORDEX catalog.  

Question: If I wanted to look at the result using your STAC browser, how would I do that?

Note that `test_export` seems to hang when I run it locally.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@
 
 * Include `PYESSV_ARCHIVE_HOME` environment variable in Dockerfile.
 * Remove redundant CLI arguments.
+* Fix bug in `THREDDSLoader` iterator introduced in 0.8.0. 
+* Remove `log_debug` option from CORDEX-CMIP6_Ouranos' runner. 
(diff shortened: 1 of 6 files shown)


fmigneault (commented):
Corresponding issues mentioned on #104

Can be edited only in the other PR and close this one directly, or partial edits here followed by extra ones in the other PR, as you prefer.

fmigneault (approved):

mishaschwartz (commented):
Yeah, same comment as @fmigneault... 

The fixes look like they're duplicated in the other PR so we should choose which one to implement and close the other one.

huard:
It's rather that I based the other branch on this one. My plan was to merge this one first, then update the other PR so we clearly see the marginal changes from it.

PR #104, opened 2025-07-01
Small fixes to create Items out of RDPS and HRDPS data sets

huard:
RDPS and HRDPS are the datasets used to train the spatial downscaling U-Net in the GeoConnexions project. 

There are small test samples here:
https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/testdata/HRDPS/catalog.html

This PR runs STAC item creation tests against those datasets. The only fixes needed to run the tests were
1. Add a default `create_uid` method in the THREDDSCatalogDataModel. Not doing so would mean having to create a custom extension for every dataset we want to add to the catalog. The default UID is just the location of the dataset relative to the `fileserver/` URL. 
2. Change how empty vertical spatial extents are encoded to match the schema. Previously `[", "]`, now `[None, None]`. 

## Notes

In the DataCube 2.2.0 schema, the `vertical_spatial_dimension` has a required `extent` defined as 
```json
extent_open": {
      "type": "array",
      "minItems": 2,
      "maxItems": 2,
      "items": {
        "type": [
          "number",
          "null"
        ]
      }
    }
```

### DIFF
--- a/STACpopulator/extensions/thredds.py
+++ b/STACpopulator/extensions/thredds.py
@@ -209,6 +209,13 @@ class THREDDSCatalogDataModel(BaseSTAC):
         data["thredds"] = THREDDSHelper(data["data"]["access_urls"])
         return data
 
+    def create_uid(self) -> str:
+        """Return a unique ID from the server location.
+
+        For datasets with a DRS, it might might more sense to use the dataset's metadata instead."""
+        location = self.data["access_urls"]["HTTPServer"].split("/fileServer/")[1]
+        return location
+
 
 # TODO: Validate services links exist ?
 # @field_validator("access_urls")
--- a/STACpopulator/input.py
+++ b/STACpopulator/input.py
@@ -103,7 +103,7 @@ class THREDDSLoader(GenericLoader):
                 url = f"{current_catalog.catalog_url}?dataset={dataset.id}"
                 yield item_name, url, attrs
 
-            for ref in current_catalog.catalog_refs:
+            for name, ref in current_catalog.catalog_refs.items():
                 catalogs.put(ref.follow())
 
     def __getitem__(self, dataset: str) -> dict:
--- a/tests/test_rdps.py
+++ b/tests/test_rdps.py
@@ -1,0 +1,21 @@
+"""Test RDPS and HRDPS netCDF files."""
+import json
+from STACpopulator.extensions.thredds import THREDDSCatalogDataModel
+
+def test_rdps():
+    attrs = json.load(open("tests/data/rdps.json"))
+    model = THREDDSCatalogDataModel.from_data(attrs)
+    item = model.stac_item()
+    assert set(model._helpers) == {"thredds", "datacube"}
+
+
+def test_hrdps():
+    attrs = json.load(open("tests/data/hrdps_sfc.json"))
+    model = THREDDSCatalogDataModel.from_data(attrs)
+    item = model.stac_item()
+    assert set(model._helpers) == {"thredds", "datacube"}
+
+    attrs = json.load(open("tests/data/hrdps_p_tt.json"))
+    model = THREDDSCatalogDataModel.from_data(attrs)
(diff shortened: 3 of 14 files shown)


tests/test_rdps.py
fmigneault, line 1: Is there a specific datacube variable or dimension that are known to be only defined in RDPS, and which we can test that they are present to ensure parsing was successful from the NetCDF?

STACpopulator/extensions/thredds.py
```diff
@@ -209,6 +209,13 @@ def thredds_helper(cls, data: dict) -> dict:
         data["thredds"] = THREDDSHelper(data["data"]["access_urls"])
         return data
 
+    def create_uid(self) -> str:
+        """Return a unique ID from the server location.
+
+        For datasets with a DRS, it might might more sense to use the dataset's metadata instead."""
+        location = self.data["access_urls"]["HTTPServer"].split("/fileServer/")[1]
+        return location
```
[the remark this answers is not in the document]
huard: This indeed determines the top-most ID field. In the parent class, `create_uid` is an `abstractmethod`, so it was left undefined here. As a result, using this class no ID was generated and the item creation failed. 

This oversight did not cause issues previously because items were created with other extensions (e.g. CMIP6, CORDEX) that defined their IDs using their own particular CV. 

This change makes it possible to create STAC items without defining another extension.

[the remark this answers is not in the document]
huard: The others inherit from this one, so they can override the method. 
```
class Cordex6DataModel(THREDDSCatalogDataModel):
```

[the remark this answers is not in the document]
huard: Yes. Ok, I ccould replace them with `__`  ?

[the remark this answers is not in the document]
fmigneault: Yes I think that works.

STACpopulator/extensions/thredds.py
```diff
@@ -209,6 +209,14 @@ def thredds_helper(cls, data: dict) -> dict:
         data["thredds"] = THREDDSHelper(data["data"]["access_urls"])
         return data
 
+    def create_uid(self) -> str:
+        """Return a unique ID from the server location.
+
+        For datasets with a DRS, it might might more sense to use the dataset's metadata instead.
+        """
+        location = self.data["access_urls"]["HTTPServer"].split("/fileServer/")[1]
```
mishaschwartz, line 217: If a dataset doesn't have an "HTTPServer" access url then can we use something else instead?

STACpopulator/input.py
```diff
@@ -103,7 +103,7 @@ def __iter__(self) -> Iterator[Tuple[str, str, MutableMapping[str, Any]]]:
                 url = f"{current_catalog.catalog_url}?dataset={dataset.id}"
                 yield item_name, url, attrs
 
-            for ref in current_catalog.catalog_refs:
+            for name, ref in current_catalog.catalog_refs.items():
```
mishaschwartz, line 106:
```suggestion
            for ref in current_catalog.catalog_refs.values():
```

STACpopulator/extensions/thredds.py
huard: Yes, which cases should we cover here? 

I assume at least either of HTTPServer or OPeNDAP would be available.

mishaschwartz: The problem is that thredds doesn't guarantee that any service type will be available. A thredds server could even just have their own custom service and none of the ones we expect.

A safer assumption is that there is at least one service (so at least one access_url) and then use the uuid library to create a unique ID using that access url as a seed:

```python
for key in self.data["access_urls"]:
    hash_m = hashlib.md5()
    hash_m.update(key.encode("utf-8"))
    return uuid.UUID(hash_m.hexdigest())
raise Exception("No access urls found!") # or just log an error or something
```

PR #105, opened 2025-07-28
extra item and collection parsers

### LINKED ISSUE #95: Add interface for additional extensions
For example, extensions can add additional links to items or collections based (like the source link used by cowbird to align permissions for stac items with THREDDS assets)

See discussion here: https://github.com/crim-ca/stac-populator/pull/88#discussion_r2128900401

mishaschwartz:
Resolves #95

### DIFF
--- a/STACpopulator/populator_base.py
+++ b/STACpopulator/populator_base.py
@@ -1,11 +1,12 @@
 import functools
+import importlib
 import inspect
 import json
 import logging
 import os
 from abc import ABC, abstractmethod
 from datetime import datetime
-from typing import Any, Dict, List, MutableMapping, Optional, Type, Union
+from typing import Any, Callable, Dict, List, MutableMapping, Optional, Type, Union
 
 import pystac
 from requests.sessions import Session
@@ -16,6 +17,7 @@ from STACpopulator.api_requests import (
     stac_host_reachable,
     stac_version_match,
 )
+from STACpopulator.exceptions import FunctionLoadError
 from STACpopulator.input import GenericLoader
 from STACpopulator.models import AnyGeometry
 from STACpopulator.stac_utils import load_config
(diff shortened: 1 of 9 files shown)


STACpopulator/populator_base.py
```diff
@@ -55,6 +68,28 @@ def __init__(
         LOGGER.info(f"Collection {self.collection_name} is assigned ID {self.collection_id}")
         self.create_stac_collection()
 
+    @staticmethod
+    def _load_extra_parser(func_str: str, extra_kwargs: dict[str, str]) -> Callable:
+        if ":" in func_str:
+            mod, func = func_str.split(":", 1)
+            try:
+                function_ns = importlib.import_module(mod)
```
fmigneault, line 76: I think this will fail unless it is a direct file (same PWD module) or that it is placed in the sys path before (as shown in the tests with `sys.path.append(str(request.path.parent / "extra_functions"))`. 

Since the original intent is to add Magpie/Cowbird extra metadata, I expect those functions would be placed is some Python files at various locations either in a complex dir hierarchy or in installed nested-module. It is better if this resolution can be done automatically without need from the user to patch the paths. 

This is how I've dealt with these combinations in Weaver: https://github.com/crim-ca/weaver/blob/8bf64a83ca894ea88dffd5e90e2b63ec3d5b268c/weaver/utils.py#L1060

fmigneault:
@huard 

> I'm not sure I understand the rationale for this

This is needed to reintroduce DACCS/PAVICS-specific features, such as including the Magpie ressource link such that Cowbird can synchronize some permissions between STAC items/assets and corresponding THREDDS entries. 

Those were removed in recent commits to make the tool more generic to other platforms. 

While using a plugin/extension could be one way to handle things, their definition can be somewhat overkill and tedious for only 1-2 extra fields. The feature can also be used to modify STAC definitions without actually introducing any extension-specific fields, such as padding missing common metadata properties.

PR #107, opened 2025-07-30
Make sure that all tests will pass even when not connected to the internet

mishaschwartz:


tests/cassettes/test_cmip6_datacube/test_datacube_helper.yaml
```diff
@@ -0,0 +1,250 @@
...
+      - close
+      Host:
+      - stac-extensions.github.io
+      User-Agent:
+      - Python-urllib/3.13
+    method: GET
+    uri: https://stac-extensions.github.io/datacube/v2.2.0/schema.json
+  response:
+    body:
+      string: "{\n  \"$schema\": \"http://json-schema.org/draft-07/schema#\",\n  \"$id\":
```
fmigneault, line 15: Is there any way to load this content from another JSON file or have it directly embedded as JSON with newlines to improve readability?

huard (approved):
Doesn't this add a new dependency that should be included somewhere ?

Please add a few lines in the developer doc to explain what this does and how to use it.

mishaschwartz:
> Doesn't this add a new dependency that should be included somewhere ?

No, we are already using pytest-recorder elsewhere in this project. It just wasn't being used consistently.

fmigneault (approved):

PR #121, opened 2025-11-11
update pre-commit with additional checks

fmigneault:


.pre-commit-config.yaml
```diff
@@ -1,4 +1,24 @@
+default_language_version:
+  python: python3
+default_stages: [commit, push]
```
mishaschwartz: This produces a warning: 

```
[WARNING] top-level `default_stages` uses deprecated stage names (commit, push) which will be removed in a future version.  run: `pre-commit migrate-config` to automatically fix this.
```

PR #123, opened 2025-11-20
bump to version 0.12.0

mishaschwartz:


fmigneault:
I've merged it as is, but next time I think it is better to do the bump directly on master. Having the tag created before master integrates it does not make the docker build `latest` update with the tagged version.

PR #124, opened 2025-11-26
Refactoring code structure

### LINKED ISSUE #117: Refactor code and improve documentation for ease of onboarding
I put here suggestions for code refactoring and improvements to facilitate onboarding.

**1. Standardize implementations using the base extension class** 
For instance, cmip6 is still boilerplate and can mislead on the right way of adding an implementation 

**2. Move `pystac-like extensions`, `helpers`, and `datamodels` into separate modules.**
This separation will improve code readibility and enable better understanding of the different layers of class interactions.
One can also more easily add and evolve pystac-like extensions this way, with clear separation from populator's helpers.

**3. Document adding a new implementation into the existing code base.**
Write comprehensive, tutorial-style documentation (inspired by [PySTAC’s example](https://pystac.readthedocs.io/en/stable/tutorials/adding-new-and-custom-extensions.html)) to guide contributors through adding new implementations

**4. Refactor implementations to reduce code repetition**
Implementations `add_[IMPL].py` currently look alike so there might a possibility to create a base implementation to reduce redundance. Also explore migrating from `argparse` to `click` that provides decorative and cleaner way for creating CLI commands.

henriaidasso:
Resolves #117 

This PR includes only structure refactoring changes (i.e., no implementation have been modified).

The refactoring follows the `extensions`, `helpers`, `datamodels`, and `populators` approach currently described in the [documentation](https://github.com/crim-ca/stac-populator/blob/master/docs/populators.md) for consistency. I proceeded by creating a seperate module for each class of the aforementioned types, then updated related imports to ensure there no regression (i.e., all tests pass).

This would greatly facilitate code navigation moving forward, since complexity will only increases from here.

Additional suggested changes:

1. Use package-level re-exporting
Avoids multiple imports and simplify usage: ```from helpers import Helper, CFHelper``` instead of
```
from helpers.base import Helper
from helpers.cf import CFHelper
```

2. Create `utils` for general utilities (e.g., logging), `api` for http requests modules, and `cli` for CLI modules.
I would need assistance on what goes where.

3. Update documentation refs accordingly

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ exclude = [
 ]
 
 [tool.setuptools.package-data]
-STACpopulator = ["**/collection_config.yml", "extensions/schemas/**/*.json"]
+STACpopulator = ["**/collection_config.yml", "helpers/schemas/**/*.json"]
 
 
 


mishaschwartz (commented):
Before we do this can we merge #122 first?
There are similar changes to the implementations in that one that will affect this PR as well. 
@huard are you able to review #122 so that we can merge that one so that this PR can build off of that one?

fmigneault (changes requested):
I agree splitting up the helpers/extensions as applicable can make it clearer (since they are reusable components), but I think separating the datamodels from their implementations is counter-productive since they are specifically designed to work with them. Maybe the only exception would be THREDDS that is reusable, but all others are very specific to their respective datasets/structure and populator invocation. 

I do not see the advantage of the "implementations->populators" moves. 
It actually seems to deduplicate most of the references across the two modules (note: I am also considering https://github.com/crim-ca/stac-populator/pull/122 that already includes a simplification & separation of THREDDS and a base populator).

PR #143, opened 2026-04-14
add filtered collection ID export

### LINKED ISSUE #101: Log progress for export
When running `stac-populator export` (eg: via https://github.com/bird-house/birdhouse-deploy/blob/2.16.3/birdhouse/components/stac/backup#L17-L28), the resulting operation does not report anything for a really long time if the referenced API contains a lot of collections/items. 

It would be good to add at least a few logs in the operation to hint toward some kind of progress.

As per https://github.com/crim-ca/stac-populator/issues/100, it seems logs are working, but only report `WARNING` and above by default. Probably `INFO` should be made the default in `stac-populator export`? Otherwise (or as well), a `--log-level INFO` might be relevant when invoked in https://github.com/bird-house/birdhouse-deploy/blob/2.16.3/birdhouse/components/stac/backup#L17-L28

fmigneault:
- Allows specifying a collection ID or regex to limit how many collections and their items are extracted from a possibly large catalog. 
- Add explicit log entries of processed or skipped Catalogs/Collections (fixes #101).
- Fix `export` target directory parent(s) creation if missing.

### DIFF
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -72,15 +72,22 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     export_parser.add_argument(
         "--ignore-duplicate-ids",
         action="store_true",
-        help="Do not raise an error if STAC items with the same ids are found in a collection.",
+        help="Do not raise an error if STAC Items with the same IDs are found in a collection.",
+    )
+    export_parser.add_argument(
+        "--collection-ignore-duplicate-ids",
+        action="store_true",
+        help="Do not raise an error if STAC Collections with the same IDs are found catalog(s).",
     )
     export_parser.add_argument(
         "-c",
         "--collection",
         nargs="*",
         help=(
-            "Only export collections whose ID matches the regex pattern. "
-            "A plain ID without pattern will only match the exact collection."
+            "Only export collections whose ID matches the entire regex pattern. "
+            "A plain ID without pattern will only match the exact collection. "
+            "If a search pattern is needed, employ adequate regexes as needed. "
+            "Multiple collection IDs can be provided. "
         ),
     )
 
@@ -125,6 +132,7 @@ def run(ns: argparse.Namespace) -> int:
                     session,
                     ns.resume,
                     ns.ignore_duplicate_ids,
+                    ns.collection_ignore_duplicate_ids,
                     ns.collection,
                 )
                 or 0
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -74,6 +74,11 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
         action="store_true",
         help="Do not raise an error if STAC items with the same ids are found in a collection.",
     )
+    export_parser.add_argument(
+        "-c", "--collection",
+        type=str,
+        help="Specific ID or regex pattern for which to filter matched collections from the catalog(s).",
+    )
 
 
 @functools.cache
@@ -109,7 +114,14 @@ def run(ns: argparse.Namespace) -> int:
         elif ns.command == "update_collection":
             return update_api_collection(ns.mode, ns.stac_collection_uri, ns.exclude_summary) or 0
         else:
-            return export_catalog(ns.directory, ns.stac_host, session, ns.resume, ns.ignore_duplicate_ids) or 0
+            return export_catalog(
+                ns.directory,
+                ns.stac_host,
+                session,
+                ns.resume,
+                ns.ignore_duplicate_ids,
+                ns.collection,
+            ) or 0
 
 
 def main(*args: str) -> int:
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -77,6 +77,7 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     export_parser.add_argument(
         "-c",
         "--collection",
+        nargs="*",
         help=(
             "Only export collections whose ID matches the regex pattern. "
             "A plain ID without pattern will only match the exact collection."
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -67,7 +67,7 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     )
     export_parser = commands_subparser.add_parser("export", description="Export a STAC catalog to JSON files on disk.")
     export_parser.add_argument("stac_host", help="STAC API URL")
-    export_parser.add_argument("directory", type=str, help="Path to a directory to write STAC catalog contents.")
+    export_parser.add_argument("directory", help="Path to a directory to write STAC catalog contents.")
     export_parser.add_argument("-r", "--resume", action="store_true", help="Resume a partial download.")
     export_parser.add_argument(
         "--ignore-duplicate-ids",
@@ -77,8 +77,10 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     export_parser.add_argument(
         "-c",
         "--collection",
-        type=str,
-        help="Specific ID or regex pattern for which to filter matched collections from the catalog(s).",
+        help=(
+            "Only export collections whose ID matches the regex pattern. "
+            "A plain ID without pattern will only match the exact collection."
+        ),
     )
 
 
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -77,7 +77,7 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
     export_parser.add_argument(
         "--collection-ignore-duplicate-ids",
         action="store_true",
-        help="Do not raise an error if STAC Collections with the same IDs are found catalog(s).",
+        help="Do not raise an error if STAC Collections with the same IDs are found in catalog(s).",
     )
     export_parser.add_argument(
         "-c",
--- a/STACpopulator/cli.py
+++ b/STACpopulator/cli.py
@@ -75,7 +75,8 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
         help="Do not raise an error if STAC items with the same ids are found in a collection.",
     )
     export_parser.add_argument(
-        "-c", "--collection",
+        "-c",
+        "--collection",
         type=str,
         help="Specific ID or regex pattern for which to filter matched collections from the catalog(s).",
     )
@@ -114,14 +115,17 @@ def run(ns: argparse.Namespace) -> int:
         elif ns.command == "update_collection":
             return update_api_collection(ns.mode, ns.stac_collection_uri, ns.exclude_summary) or 0
         else:
-            return export_catalog(
-                ns.directory,
-                ns.stac_host,
-                session,
-                ns.resume,
-                ns.ignore_duplicate_ids,
-                ns.collection,
-            ) or 0
+            return (
+                export_catalog(
+                    ns.directory,
+                    ns.stac_host,
+                    session,
+                    ns.resume,
+                    ns.ignore_duplicate_ids,
+                    ns.collection,
+                )
+                or 0
+            )
 
 
 def main(*args: str) -> int:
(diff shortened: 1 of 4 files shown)


STACpopulator/cli.py
```diff
@@ -74,6 +74,12 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
         action="store_true",
         help="Do not raise an error if STAC items with the same ids are found in a collection.",
     )
+    export_parser.add_argument(
+        "-c",
+        "--collection",
+        type=str,
+        help="Specific ID or regex pattern for which to filter matched collections from the catalog(s).",
```
mishaschwartz: I'm a little confused by the wording here. What about:

```suggestion
        help="Only export collections whose id matches this regex.",
```

STACpopulator/cli.py
```diff
@@ -74,6 +74,12 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
         action="store_true",
         help="Do not raise an error if STAC items with the same ids are found in a collection.",
     )
+    export_parser.add_argument(
+        "-c",
+        "--collection",
+        type=str,
```
mishaschwartz: This is the default type so we don't need to specify it.

STACpopulator/export.py
```diff
@@ -153,11 +164,15 @@ def export_catalog(
...
     """Export a STAC catalog to files on disk."""
     stac_api_io = StacApiIO()
     stac_api_io.session = session
     directory = pathlib.Path(directory)
     client = Client.open(stac_host, stac_io=stac_api_io)
     start_time = time.time()
-    _export_catalog(client, directory, start_time, resume, ignore_duplicate_ids)
+    collection_pattern = collection
+    if isinstance(collection, str):  # convenience plain ID to regex
+        collection_pattern = re.compile(f"^{collection.strip('^$')}$")
```
mishaschwartz: This forces the regex to match the whole string which may not be what the user actually wants.
I think we should respect the regex that the user actually submits and not try to add more to it. 

So if the user submits `abc` then that means something different than `^abc` or `^abc$`.

Or if we actually want to convert all of those to `^abc$` then we need to be very explicit about that in the documentation and the help message.

STACpopulator/export.py
```diff
@@ -144,7 +146,16 @@ def _export_catalog(
             ignore_duplicate_ids,
         )
     for child in client.get_children():
-        _export_catalog(child, directory, start_time, resume, ignore_duplicate_ids)
+        if (
+            isinstance(child, pystac_client.CollectionClient)
+            and collection_pattern
+            and not collection_pattern.match(child.id)
```
mishaschwartz: We should use `.search` instead of `.match` if we don't force `^` and `$` as I suggest in the other comment.

STACpopulator/cli.py
```diff
@@ -74,6 +74,12 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
         action="store_true",
         help="Do not raise an error if STAC items with the same ids are found in a collection.",
     )
+    export_parser.add_argument(
```
mishaschwartz, line 82: Another way to avoid the whole question of whether to add ^ and $ to the regex but still allow users to specify multiple collections to export is to:

- don't treat this as a regex but as a string that has to be matched directly
- add the `nargs="*"` option here to allow the user to specify multiple collection names
- when we encounter a new collection we can just do `if child.id in collection` (where collection is a list of id names)

STACpopulator/export.py
```diff
@@ -144,7 +146,16 @@ def _export_catalog(
...
+        ):
+            LOGGER.info(f"Skipping collection {child.id} as it does not match the collection pattern")
+            continue
+        child_type = "collection" if isinstance(child, pystac_client.CollectionClient) else "catalog"
```
mishaschwartz, line 174: In the case where there are multiple collections so there is a more recursive structure we could have multiple collections with the same name. For example:

- catalog1
  - catalog2
    - collection1 
  - catalog3
    - collection1

So now we have two collections named "collection1". With the current change there is no way to differentiate between the two. There is now the possibility to have nested catalogs for the stac implementation that we use (https://github.com/stac-utils/stac-fastapi-pgstac/issues/79) so I think we have to consider this.

What if we went with the solution without the regex (see other comment) and require that the collection name actually be a path from the top level collection. So if I wanted to specify only the first `collection1` only I'd say: `--collection /catalog1/catalog2/collection1`

We could even support relative and absolute paths so we could also say `--collection catalog2/collection1` because the `stac_host` parameter would already be pointing to the base catalog named `catalog1`

STACpopulator/cli.py
fmigneault: The wording was to indicate that it could be a literal ID or a regex. However, I realize now that there might be side effects, like "my-collection" not liking the `-` as range. I'll adjust as you propose so it is clearer that it is employed as regex either way.

STACpopulator/export.py
fmigneault: The thing is that most of the time, `abc` will not match anything unless some `.*` or similar prefix/suffix are explicitly added, or that the name is exactly that, so it behaves the same way as `^abc$`. 

My goal is to make the parameter as natural as possible. Eg: if I indicate `-c climate`, I would like to match that collection without having to explicitly write `^climate$` and I would rather not match every `"...climate..."` collections. If I want to multi-match, it is natural for the input to look like a pattern, eg.: `climate-.*`.  The reason I added the `^$` is that it wasn't very clear that "`climate`" was an exact match.

STACpopulator/export.py
fmigneault: That's the thing. I don't want a search since it is very common to search just for one collection, and it is more natural to say `-c desired-collection`, since `^$` will require extra quote to avoid shell extension.

STACpopulator/cli.py
fmigneault: I like the `narg` idea because it is easier to do `-c collectionA -c collectionB` than writing its regex. However, I think the same pattern/literal ID simplification should apply to them.

STACpopulator/export.py
fmigneault: While it is true that the Catalogs-spec is implemented to allow nesting, as far as I am aware, the pystac-client currently only still do a search based on `/search` across all collections. The pystac-client has "nested catalog" support for "[static catalog](https://github.com/radiantearth/stac-spec/blob/master/best-practices.md#static-and-dynamic-catalogs)" dir/files with different base/root "URI", but this is not really our case of API dynamic catalog since we target a `stac_host` and it does not yet implement catalog endpoints.

Also, this extension defines the catalogs as graph nodes, so they would not necessarily be "nested" like a file structure, but could be exported with `catalogX/collectionY` as flat list. From the DB perspective, `collection1` could be literally the same. 

I am participating in the design of the [StacLabs/stac-fastapi-catalogs-extension](https://github.com/StacLabs/stac-fastapi-catalogs-extension). I propose we revisit once it is actually implemented to improve this feature as needed.

mishaschwartz: Pystac-client does actually use recursive search for both catalogs and APIs (I raised an issue about this a while ago here: https://github.com/stac-utils/pystac-client/issues/798).

> Also, this extension defines the catalogs as graph nodes

Ok that's fine but we're talking about how people will use the `--collections` argument for stac-populator and the implementation of the backend shouldn't matter. People want to know that when they specify a collection it won't match more or fewer collections than they expect.

STACpopulator/export.py
mishaschwartz: So then just say that `-c` will match the collection id exactly and avoid all the potential confusion of a regex.

STACpopulator/export.py
mishaschwartz: Ok but the whole point of the regex is to potentially match more than one collection at a time.

If I give any of the following regexes:

- `^abc|def$`
- `abc|def`
- `^abc$|^def$`
- `abc$|^def`

Then some of these will match exactly the same pattern as others and others will be slightly different in subtle and non-obvious ways.

If the point is to only match one collection ever then just say that and handle that in the code. If the idea is to allow users to specify multiple collections then say that and allow for multiple `-c` arguments (using `nargs`).

STACpopulator/export.py
fmigneault: Well, the way that it handles it is that, if item search is supported (ie: an API providing `/search`), it ignores the recursive option entirely since it always (or never depending on your point of view) applies, since all items are returned at flat list across all catalog/collections. The recursive portion only comes from the static side using a different root `catalog.json` file, but each one is then processed on its own with the fallback get-items.

According to how the backend stores the collections (at least currently with pgstac), they are all within the same table. There is no concept of nesting. Most probably, when `/catalogs/...` endpoints will be added, they will only give the "impression" of some hierarchical structure if the collection graph nodes are ordered this way, but you could technically have children participating under multiple parent catalogs. To do so, the collections would still rely on a unique ID. Therefore, export would work as flat list, and search could be done throughout all the catalogs. 

What could be done is for `pystac-client` to itself do a precheck of `conforms_to(catalogs-extension)`, which would then yield results adequately based on a catalog prefix. We could then to `collection_pattern.rsplit("/", 1)` or similar to obtain that prefix and pass it down to `search()`, but we need to see how it actually gets integrated before doing so.

STACpopulator/export.py
fmigneault: I allow both variants based on the natural interpretation of the value. If you mention `-c someCollection`, it is obvious you mean you want that specific one. If you provide explicit regex elements like `-c '.*-climate'`, it is also obvious that you are searching for many. Both can be converted to a regex pattern for convenience.

STACpopulator/export.py
fmigneault: The latest changes allow both, so you can do `-c test -c *.other` or `-c '^test$|^.*other$'`. 

I don't mind removing the `^$` explicitly added to leave the flexibility to control those if provided directly, but then I have no choice to keep using `match` since `search` could yield many results when the user only indents to find one. 

My aim is that users that want something specific don't have to start thinking about regexes just to "undo" a pattern search. It is much more common to only look for a specific collection, or to look a specific pattern matching similar concepts, but very unusual to need convoluted combinations of unrelated IDs.

STACpopulator/cli.py
```diff
@@ -67,12 +67,28 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
...
     export_parser.add_argument(
         "--ignore-duplicate-ids",
         action="store_true",
-        help="Do not raise an error if STAC items with the same ids are found in a collection.",
+        help="Do not raise an error if STAC Items with the same IDs are found in a collection.",
+    )
+    export_parser.add_argument(
+        "--collection-ignore-duplicate-ids",
+        action="store_true",
+        help="Do not raise an error if STAC Collections with the same IDs are found catalog(s).",
```
mishaschwartz: typo: "found in catalog(s)"

STACpopulator/cli.py
```diff
@@ -67,12 +67,28 @@ def add_parser_args(parser: argparse.ArgumentParser) -> None:
...
+    )
+    export_parser.add_argument(
+        "-c",
+        "--collection",
+        nargs="*",
+        help=(
+            "Only export collections whose ID matches the entire regex pattern. "
```
mishaschwartz, line 87: 👍

STACpopulator/export.py
```diff
@@ -143,8 +149,39 @@ def _export_catalog(
...
+                f"the collection patterns {json.dumps([col.pattern for col in collection_patterns])}."
+            )
+            continue
+        if found_cols & {child.id}:
+            if collection_ignore_duplicate_ids:
+                LOGGER.warning(f"Duplicate collection ID {child.id} was already processed. Skipping it.")
+                continue
+            else:
+                raise DuplicateIDError(
+                    f"Duplicate collection ID {child.id} was already processed and were not allowed. "
```
mishaschwartz: typo: "and was not allowed"

mishaschwartz (approved):
Thank you that makes more sense to me now.
This looks good pending a few small typos in log messages and help strings:

PR #145, opened 2026-04-23
fix interaction between pre-commit hooks and bump-my-version

mishaschwartz:
When bump-my-version runs and creates a new version it adds an additional newline to the CHANGES.md file. This line is then flagged by markdownlint when then pre-commit hooks run and the commit should be blocked.

However, since version bump commits are sometimes pushed directly to master instead of through a PR then these checks don't run. This means that the next PR **after** the version has been updated will fail the pre-commit checks. 

For example, see #144 which was the first commit after https://github.com/crim-ca/stac-populator/commit/c5a02c7568968bd5af2cfc94d1e027cc458e7c62 was pushed to master. This caused this action to fail (https://github.com/crim-ca/stac-populator/actions/runs/24754811983/job/72425655164)

In order to fix this, this PR updates the bump-my-version configuration in pyproject.toml in order to not include the newline. This temporarily fixes the issue but the better solution would be to ensure that the checks run by github actions run every time there is a change to this repo. In order to do this we should disable the ability to push directly to master. This means that every change has to go through a PR, including version bumps. This will ensure that all changes are flagged if an action fails.
I don't have permission to do this so could @fmigneault (or someone else with permission) please update that when you can.

Some additional changes made in this PR in order to update the pre-commit hooks as well:

- added the `--ignore-path` flag to markdownlint so that it doesn't try to fix non-project files
- added `venv/` to the .gitignore file since that is a common name for a virtual environment (and so that markdownlint now ignores it as expected)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased](https://github.com/crim-ca/stac-populator) (latest)
 
-* fix interaction between pre-commit hooks and bump-my-version.
+* Fix interaction between pre-commit hooks and bump-my-version.
 
 ## [0.15.0](https://github.com/crim-ca/stac-populator/tree/0.15.0) (2026-04-21)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,10 +2,9 @@
 
 ## [Unreleased](https://github.com/crim-ca/stac-populator) (latest)
 
-<!-- insert list items of new changes here -->
+* fix interaction between pre-commit hooks and bump-my-version.
 
 ## [0.15.0](https://github.com/crim-ca/stac-populator/tree/0.15.0) (2026-04-21)
-
 
 * Add `-c`/`--collection` option to `export` operation to filter which collections should be exported,
   rather than the entire catalog. The value can be the plain Collection ID or a regex pattern.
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -91,8 +91,7 @@ replace = """
 
 <!-- insert list items of new changes here -->
 
-## [{new_version}](https://github.com/crim-ca/stac-populator/tree/{new_version}) ({now:%Y-%m-%d})
-"""
```
fmigneault, line 95: Could the change instead be that `search` is made multiline so it search/replace the same way including newlines?

CHANGES.md
```diff
@@ -2,11 +2,10 @@
 
 ## [Unreleased](https://github.com/crim-ca/stac-populator) (latest)
 
-<!-- insert list items of new changes here -->
+* fix interaction between pre-commit hooks and bump-my-version.
```
fmigneault: casing of "Fix"

fmigneault (commented):
Nicely spotted.

I think forcing a PR just to bump will add unnecessary overhead because the tags will then have to be re-adjusted to match the merged one on master, and then re-released.

I'm fine with this fix if it makes it consistent between the bumped release and following PRs/checks. Normally, anything else than the bump would have been pre-validated by the previous PR.

fmigneault (approved):

